PR #2, opened 2021-06-05
Interview

alexz429:
Adds interviewer signup and listing functionalities


kallentu (changes requested):
Ack, there's a lot of linter issues, can you fix these and I'll take a skim over.

src/components/db.ts
```diff
@@ -10,7 +10,9 @@ export const openDB = async (): Promise<Database> => {
       filename: 'db/bot.db',
       driver: sqlite3.Database
     });
+    //initialize all relevant tables
     await db.run('CREATE TABLE IF NOT EXISTS saved_data (msg_id INTEGER PRIMARY KEY,data TEXT NOT NULL);');
+    await db.run('CREATE TABLE IF NOT EXISTS Interviewers (UserId TEXT, Link TEXT)');
```
gzcharleszhang, line 15: we should prob make `UserId` the primary key and add not null constraint for `Link`. Also be mindful of the column name styles here, for `saved_data` you are using snake case, but for `Interviewers` you are using camel case. We should just stick with one, personally, I've always done snake case for SQL stuff. So `interviewers` table with columns `user_id` and `link`

src/components/interview.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+function parseLink(link: string) {
+  if (link.includes('calendly.com') || link.includes('x.ai')) {
+    if (!link.startsWith('http')) {
+      link = 'https://' + link;
+    }
+    return link;
+  } else {
+    return undefined;
```
gzcharleszhang, line 16: I don't think returning undefined is good practice, maybe return `null`?

src/components/interview.ts
```diff
@@ -0,0 +1,90 @@
...
+    count++;
+    console.log(rows.UserId.toString());
+    listString +=
+      '**' + (await client.users.fetch(rows['UserId'].toString())).tag + '** | [Calendar](' + rows['Link'] + ')\n\n';
+  }
+  outEmbed.setDescription(listString);
+  await message.channel.send(outEmbed);
+}
+
+function shuffleArray(array: Interviewer[]) {
```
gzcharleszhang, line 84: let's use lodash for things like this https://lodash.com/docs/#shuffle

gzcharleszhang:
```
import _ from 'lodash';

_.shuffle(something)
```

src/components/interview.ts
```diff
@@ -0,0 +1,90 @@
...
+    case 'list':
+      listInterviewers(message, client);
+      break;
+    default:
+      help(message);
+      break;
+  }
+}
+
+export async function help(message: Discord.Message) {
```
gzcharleszhang, line 35: do we need to export this function?

src/components/interview.ts
```diff
@@ -0,0 +1,90 @@
...
+    help(message);
+    return;
+  }
+  const parsedLink = parseLink(link);
+  if (!parsedLink) {
+    message.channel.send(`Hmmm... I don't seem to recognize your meeting link. Be sure to use calendly or x.ai.`);
+    return;
+  }
+  const res = await db.get('SELECT * FROM Interviewers WHERE UserID = ?', id);
+  if (res == undefined) {
```
gzcharleszhang, line 57: can we do `!res` here or do we have to explicitly check for `undefined`

src/components/db.ts
alexz429: Should I use a String as primary key though? I was worried about performance issues. Fixing everything else in a bit.

gzcharleszhang: I think having a primary key/index is better than not having one. If you are concerned about integer comparisons vs string comparisons, I think it's fine because we won't have a lot of rows, and the fact that we don't have to maintain our own integer primary key as well as discord user ID foreign key outweighs the slight performance decrease.

gzcharleszhang (changes requested):
some comments

src/components/db.ts
gzcharleszhang: Actually isn’t discord IDs numeral objects? I think we can use integers in sqlite3 then right?

alexz429: ID's are too big for Javascript to hold them precisely apparently. When I had them as numbers the last few digits had precision issues.

gzcharleszhang: oh ok then using text is fine

package.json
```diff
@@ -21,6 +21,7 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
+    "@types/lodash.shuffle": "^4.2.6",
```
gzcharleszhang, line 24: we can just import the entire lodash library, it'll be help for other operations as well

src/components/interview.ts
```diff
@@ -0,0 +1,92 @@
...
+  } else {
+    db.run('UPDATE interviewers SET link = ? WHERE user_id = ?', parsedLink, id);
+    message.channel.send(`<@${id}>, your info has been changed.`);
+  }
+}
+
+async function listInterviewers(message: Discord.Message, client: Discord.Client): Promise<void> {
+  const db = await openDB();
+  const res = _.shuffle(await db.all('SELECT * FROM interviewers')) as Interviewer[];
+  shuffleArray(await db.all('SELECT * FROM interviewers'));
```
gzcharleszhang, line 70: forgot to remove?

src/components/interview.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  const outEmbed = new Discord.MessageEmbed().setColor('#0099ff').setTitle('Available Interviewers');
+  let listString = '';
+  let count = 0;
+  for (const rows of res) {
+    if (count == 6) break;
```
gzcharleszhang, line 76: instead of having this magic number here, we should define a constant so it's clear what this is

src/components/interview.ts
```diff
@@ -0,0 +1,92 @@
...
+    count++;
+    console.log(rows.user_id.toString());
```
gzcharleszhang, line 78: don't forget to remove

gzcharleszhang (approved):
LGTM

src/components/interview.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+}
+
+async function listInterviewers(message: Discord.Message, client: Discord.Client): Promise<void> {
+  const db = await openDB();
+  const res = _.shuffle(await db.all('SELECT * FROM interviewers')) as Interviewer[];
+
+  const outEmbed = new Discord.MessageEmbed().setColor('#0099ff').setTitle('Available Interviewers');
+  let listString = '';
+  let count = 0;
```
kallentu, line 75: Nit: I see what you're doing with count, but you don't need to use the variable after the loop, so why not use a typical for loop to do the work?

src/components/db.ts
```diff
@@ -10,7 +10,9 @@ export const openDB = async (): Promise<Database> => {
       filename: 'db/bot.db',
       driver: sqlite3.Database
     });
+    //initialize all relevant tables
     await db.run('CREATE TABLE IF NOT EXISTS saved_data (msg_id INTEGER PRIMARY KEY,data TEXT NOT NULL);');
+    await db.run('CREATE TABLE IF NOT EXISTS interviewers (user_id TEXT PRIMARY KEY, link TEXT)');
```
kallentu, line 15: NOT NULL constraint for link if you can.

kallentu (approved):
Looks good, just a couple of nits to fix before pushing.

PR #3, opened 2021-06-05
Suggest command

marko-polo-cheno:


src/bot.ts
```diff
@@ -53,6 +55,21 @@ const handleCommand = async (message: Discord.Message, command: string, args: st
   }
 };
 
+const handleAnonCommand = async (message: Discord.Message, command: string, args: string[]) => {
+  // log command and its author info
+  logger.info({
+    event: 'command',
+    messageId: message.id,
```
marko-polo-cheno, line 62: disregarding discordjs/Commando stuff for now

src/bot.ts
```diff
@@ -53,6 +55,21 @@ const handleCommand = async (message: Discord.Message, command: string, args: st
...
+    command,
+    args
+  });
+
+  switch (command) {
+    case 'suggest':
+      suggestCmd(message, args);
```
gzcharleszhang, line 69: i think we can move this back to the normal handle command function since we are logging everything

src/commands/suggest.ts
```diff
@@ -0,0 +1,41 @@
...
+  let words = '';
+  let word = '';
+  // Turn args into suggestion
+  for (word in args) {
+    words += args[word] + ' ';
+  }
+
+  if (words == '') {
+    message.channel.send('Codey sees an empty suggestion! Try again.');
+  } else if (words == helpArg) {
```
gzcharleszhang, line 18: maybe you could just check `args[0] === "help"`

src/commands/suggest.ts
```diff
@@ -0,0 +1,41 @@
...
+      (
+        await db
+      ).run(
+        'INSERT INTO suggestions (suggestion_id, suggestion_author, suggestion, suggestion_state) VALUES(?,?,?,?);',
+        [null, message.author.id, words, state]
+      );
+
+      // Confirm suggestion was taken
+      message.channel.send('Codey has recieved your suggestion.');
+    } catch (err) {
```
gzcharleszhang, line 34: i don't think we need to try catch here if we have try catch in the handleCommand function, since this will prevent the app from logging errors, because we can send the error message in that error handler instead.

src/components/db.ts
```diff
@@ -11,6 +11,10 @@ export const openDB = async (): Promise<Database> => {
       driver: sqlite3.Database
     });
     await db.run('CREATE TABLE IF NOT EXISTS saved_data (msg_id INTEGER PRIMARY KEY,data TEXT NOT NULL);');
+    await db.run(
+      'CREATE TABLE IF NOT EXISTS suggestions (suggestion_id INTEGER PRIMARY KEY NOT NULL, suggestion_author VARCHAR(255) NOT NULL,' +
+        'created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,suggestion VARCHAR(500) NOT NULL, suggestion_state VARCHAR(1) NOT NULL);'
```
gzcharleszhang, line 16: can we use full names for state instead of a single character?

gzcharleszhang: also I think we can just leave suggestion as `TEXT` instead of setting a limit on number of chars, unless you want to check the character limit on the application level as well

src/commands/ping.ts
```diff
@@ -0,0 +1,8 @@
+// Codey ping Command
+
+import Discord from 'discord.js';
+
+export const pingCmd = async (message: Discord.Message) => {
+  message.channel.send('pong');
+  return;
```
gzcharleszhang, line 7: you don't need return here if you are not returning anything

src/commands/suggest.ts
```diff
@@ -0,0 +1,39 @@
+// Codey suggest Command
+import Discord from 'discord.js';
+import { openDB } from '../components/db';
+
+export const suggestCmd = async (message: Discord.Message, args: string[]) => {
+  const db = openDB();
+  const state = 'create';
+  const helpArg = 'help';
+  let words = '';
+  let word = '';
```
gzcharleszhang, line 10: I don't think you need to define this var here if you're doing `for (word in args)`

src/commands/suggest.ts
```diff
@@ -0,0 +1,39 @@
+// Codey suggest Command
+import Discord from 'discord.js';
+import { openDB } from '../components/db';
+
+export const suggestCmd = async (message: Discord.Message, args: string[]) => {
+  const db = openDB();
+  const state = 'create';
```
gzcharleszhang, line 7: I'd prefer if we have some kind of enum for states, since we will start adding more https://www.typescriptlang.org/docs/handbook/enums.html

src/commands/suggest.ts
```diff
@@ -0,0 +1,39 @@
...
+  const helpArg = 'help';
+  let words = '';
+  let word = '';
+  // Turn args into suggestion
+  for (word in args) {
+    words += args[word] + ' ';
+  }
+
+  console.log('!' + args[0] + '!');
```
gzcharleszhang, line 16: remove debug

src/commands/suggest.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  if (words == '') {
+    message.channel.send('Codey sees an empty suggestion! Try again.');
+  } else if (args[0].toLowerCase() === helpArg) {
+    message.channel.send(
+      '.suggest <suggestion> \nYour <suggestion> should only contain text and is capped at about 100 words.'
+    );
+  } else {
+    // Save suggestion into DB
+    (
```
gzcharleszhang, line 26: I don't think you need brackets here

src/commands/suggest.ts
```diff
@@ -0,0 +1,39 @@
+// Codey suggest Command
+import Discord from 'discord.js';
+import { openDB } from '../components/db';
+
+export const suggestCmd = async (message: Discord.Message, args: string[]) => {
+  const db = openDB();
```
gzcharleszhang, line 6: you should do `const db = await openDB()`

src/commands/suggest.ts
```diff
@@ -0,0 +1,39 @@
...
+    message.channel.send('Codey sees an empty suggestion! Try again.');
+  } else if (args[0].toLowerCase() === helpArg) {
+    message.channel.send(
+      '.suggest <suggestion> \nYour <suggestion> should only contain text and is capped at about 100 words.'
+    );
+  } else {
+    // Save suggestion into DB
+    (
+      await db
+    ).run('INSERT INTO suggestions (suggestion_id, suggestion_author, suggestion, suggestion_state) VALUES(?,?,?,?);', [
```
gzcharleszhang, line 28: do you need to include `suggestion_id` if it's auto incremented?

src/components/db.ts
```diff
@@ -11,6 +11,10 @@ export const openDB = async (): Promise<Database> => {
       driver: sqlite3.Database
     });
     await db.run('CREATE TABLE IF NOT EXISTS saved_data (msg_id INTEGER PRIMARY KEY,data TEXT NOT NULL);');
+    await db.run(
+      'CREATE TABLE IF NOT EXISTS suggestions (suggestion_id INTEGER PRIMARY KEY NOT NULL, suggestion_author VARCHAR(255) NOT NULL,' +
```
gzcharleszhang, line 15: I think we can shorten these column names to just `id`, `author`, `state` and so on, since given that they're in the suggestions table, we will have enough context to understand what they are

src/commands/ping.ts
marko-polo-cheno: the linter thing keeps giving me warnings (github actions) i dunno how to fix those

gzcharleszhang: it's because you need to specify a return type, so myabe like `async (message: Discord.Message): Promise<void> => ...`

src/commands/suggest.ts
```diff
@@ -0,0 +1,47 @@
...
+  db.run('DROP TABLE suggestions');
+  const res = await db.all('SELECT * FROM suggestions');
+  for (const rows of res) {
+    console.log(rows['id'], rows['created_at'], rows['author'], rows['suggestion'], rows['state']);
+  }
+*/
+
+// All states of suggestion records
+enum SuggestionStates {
+  Create = 'Create',
```
gzcharleszhang, line 16: nit: rename to `Created` makes more sense? Also since we are using enums, the actual values could be integers.
```
enum SuggestionStates {
Created = 1,
Rejected,
Accepted,
Pending
}
```
Sorry if I was kinda misleading you, I wanted you to use the full name in the DB because there wasn't enough reference in the code to understand what the states meant. Now that we have this enum, we can just reference the it for more context, so we could just store integers in the DB.

src/commands/suggest.ts
```diff
@@ -0,0 +1,47 @@
...
+  const db = await openDB();
+
+  const helpArg = 'help';
+  let words = '';
+  // Turn args into suggestion
+  for (const word in args) {
+    words += args[word] + ' ';
+  }
+
+  if (words == '') {
```
gzcharleszhang, line 32: should probably do `===`, weird that linter didn't say anything

gzcharleszhang (approved):
Looks good after those 2 comments

src/commands/suggest.ts
```diff
@@ -0,0 +1,47 @@
...
+    message.channel.send('Codey sees an empty suggestion! Try again.');
+  } else if (args[0].toLowerCase() === helpArg) {
+    message.channel.send('.suggest <suggestion> \nYour <suggestion> should only contain text.');
```
kallentu, line 35: Can you throw the command example part into a code block? Like show it as `.suggest <suggestion>`

kallentu: We don't have to focus too much on formatting, I just want a little fidelity tis all

kallentu (approved):

PR #15, opened 2021-06-09
Create issue templates

gzcharleszhang:
Added two templates, one for bug reports and another for feature requests. Let me know if you'd add/change anything.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,5 +1,5 @@
 ---
-name: Bug report
+name: Bug Report
 about: Create a report to help us improve
 title: ''
 labels: bug
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,6 +1,6 @@
 ---
 name: Bug Report
-about: Create a report to help us improve
+about: Create a report to help us improve and fix existing issues
 title: ''
 labels: bug
 assignees: ''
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,26 @@
+---
+name: Bug report
+about: Create a report to help us improve
```
kallentu, line 3: Create a bug report to help us improve and fix reported issues.

kallentu: nit: aaaaa capitalize the Bug Report and Feature Request titles

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,20 @@
...
+title: ''
+labels: request
+assignees: ''
+
+---
+
+**Is your feature request related to a problem? Please describe.**
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
+
+**Describe the solution you'd like**
```
kallentu, line 13: Maybe something like if its a command, what is the command, how would a user use this?
Just for more specificity since otherwise we'd have to follow up on that, we can just write it outright.

kallentu (commented):
Are we also going to have a template for PRs? Or is that in the works/done? I'm unsure of the status.

gzcharleszhang:
@kallentu already merged the PR template last night

.github/ISSUE_TEMPLATE/feature_request.md
gzcharleszhang: true this is the default template that's why it's so generic, we should be more specific indeed

kallentu (approved):
gud

Issue #5, opened 2021-06-09
Mock interviews: create a command so mock interviewers can add their expertise domains.


gzcharleszhang:
A command such as `.interviewer domain frontend`, where each interviewer can be associated with an arbitrary number of domains. Available domain options should be provided by us and be limited, we shouldn't allow users to add arbitrary domains.

Issue #4, opened 2021-06-09
Mock interviews: add an option to query available interviewers by domain


gzcharleszhang:
We should start with just querying by one single domain e.g. `.interviewers list domain1`. The embed response should just contain interviewers that chose to be in that domain.

We should brainstorm available domain options, having limited domain options can help us maintain the complexity of our code and mitigate risks.

PR #19, opened 2021-06-11
Interviewer modify and query by expertise domains

alexz429:
# Related Issues
resolves https://github.com/uwcsc/codeybot/issues/5
resolves https://github.com/uwcsc/codeybot/issues/4
# Summary of Changes 
Add comments to interview.ts
Add functionality to clear interviewer data 
Add functionality to add expertise domain 
Add functionality to filter by expertise domain
Add functionality to view own calendar link and domains
# Steps to Reproduce
```
//create interviewer profile
.interviewer signup YOUR_CALENDLY_OR_XAI_LINK
//add domains
.interviewer domain [frontend/backend/pm/design]
//filter by domain
.interviewer list (frontend/backend/pm/design)
//view own profile
.interviewer profile
//clear profile
.interviewer clear
```
image
image

### DIFF
--- a/src/components/db.ts
+++ b/src/components/db.ts
@@ -11,12 +11,13 @@ export const openDB = async (): Promise<Database> => {
       driver: sqlite3.Database
     });
     //initialize all relevant tables
-    await db.run('CREATE TABLE IF NOT EXISTS saved_data (msg_id INTEGER PRIMARY KEY,data TEXT NOT NULL);');
+    await db.run('CREATE TABLE IF NOT EXISTS saved_data (msg_id INTEGER PRIMARY KEY, data TEXT NOT NULL);');
     await db.run(
       'CREATE TABLE IF NOT EXISTS suggestions (id INTEGER PRIMARY KEY NOT NULL, author VARCHAR(255) NOT NULL,' +
         'created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP, suggestion TEXT NOT NULL, state INTEGER NOT NULL);'
     );
     await db.run('CREATE TABLE IF NOT EXISTS interviewers (user_id TEXT PRIMARY KEY, link TEXT NOT NULL)');
+    await db.run('CREATE TABLE IF NOT EXISTS domains (user_id TEXT NOT NULL, domain TEXT NOT NULL)');
   }
   return db;
 };
--- a/src/components/db.ts
+++ b/src/components/db.ts
@@ -18,6 +18,7 @@ export const openDB = async (): Promise<Database> => {
     );
     await db.run('CREATE TABLE IF NOT EXISTS interviewers (user_id TEXT PRIMARY KEY, link TEXT NOT NULL)');
     await db.run('CREATE TABLE IF NOT EXISTS domains (user_id TEXT NOT NULL, domain TEXT NOT NULL)');
+    await db.run('CREATE INDEX IF NOT EXISTS ix_domains_domain ON domains (domain)');
   }
   return db;
 };
--- a/src/components/interview.ts
+++ b/src/components/interview.ts
@@ -15,16 +15,7 @@ interface Interviewer {
 }
 
 function domainString() {
-  let output = '';
-  let first_domain = true;
-  for (const i of Object.values(available_domains)) {
-    if (!first_domain) {
-      output += ', ';
-    }
-    first_domain = false;
-    output += i;
-  }
-  return output;
+  return _.join(Object.values(available_domains), ', ');
 }
 
 function parseLink(link: string) {
@@ -103,10 +94,10 @@ async function addInterviewer(message: Discord.Message, args: string[]): Promise
   const res = await db.get('SELECT * FROM interviewers WHERE user_id = ?', id);
   if (!res) {
     db.run('INSERT INTO interviewers (user_id, link) VALUES(? , ?)', id, parsedLink);
-    message.reply(`<@${id}>, your info has been added. Thanks for helping out! :codeyLove:`);
+    message.reply(`Your info has been added. Thanks for helping out! :codeyLove:`);
   } else {
     db.run('UPDATE interviewers SET link = ? WHERE user_id = ?', parsedLink, id);
-    message.reply(`<@${id}>, your info has been changed.`);
+    message.reply(`Your info has been changed.`);
   }
 }
 
--- a/src/components/interview.ts
+++ b/src/components/interview.ts
@@ -1,7 +1,6 @@
 import { openDB } from './db';
 import Discord from 'discord.js';
 import _ from 'lodash';
-import { Database, Statement } from 'sqlite3';
 
 const RESULTS_PER_PAGE = 6;
 
--- a/src/components/interview.ts
+++ b/src/components/interview.ts
@@ -1,22 +1,24 @@
 import { openDB } from './db';
 import Discord from 'discord.js';
 import _ from 'lodash';
+import { Database } from 'sqlite3';
 
 const RESULTS_PER_PAGE = 6;
+
+//maps from key to readable string
 const available_domains: { [key: string]: string } = {
   frontend: 'Frontend',
   backend: 'Backend',
   design: 'Design',
   pm: 'PM'
 };
+
 interface Interviewer {
   user_id: string;
   link: string;
 }
 
-function domainString() {
-  return _.join(Object.values(available_domains), ', ');
-}
+const getDomainString = (domains: { [key: string]: string }) => _.join(Object.values(domains), ', ');
 
 function parseLink(link: string) {
   //checks if link is (roughly) one from calendly or x.ai
@@ -67,7 +69,7 @@ async function help(message: Discord.Message): Promise<void> {
       { name: 'Add Interviewer Domain', value: '`.interviewer domain [domain]`' },
       { name: 'Show Interviewer Profile', value: '`.interviewer profile`' },
       { name: 'Clear Profile', value: '`.interviewer clear`' },
-      { name: 'Available Domains', value: '`' + domainString() + '`' }
+      { name: 'Available Domains', value: '`' + getDomainString(available_domains) + '`' }
     );
   await message.reply(response);
 }
@@ -111,7 +113,7 @@ async function listInterviewers(message: Discord.Message, client: Discord.Client
   if (!domain) {
     res = await db.all('SELECT * FROM interviewers');
   } else if (!(domain in available_domains)) {
-    await message.reply('Not a valid domain, valid domains are: ' + domainString());
+    await message.reply('Not a valid domain, valid domains are: ' + getDomainString(available_domains));
     return;
   } else {
     res = await db.all(
@@ -194,15 +196,15 @@ async function addDomain(message: Discord.Message, args: string[]): Promise<void
 
   //check if domain valid
   if (!(domain in available_domains)) {
-    await message.reply('Not a valid domain, valid domains are: ' + domainString());
+    await message.reply('Not a valid domain, valid domains are: ' + getDomainString(available_domains));
     return;
   }
 
   //check if user already in domain
-  const res = await db.get('SELECT * FROM domains WHERE user_id = ? AND domain = ?', id, domain);
+  const inDomain = await db.get('SELECT * FROM domains WHERE user_id = ? AND domain = ?', id, domain);
 
   //toggles on/off user's domain
-  if (!res) {
+  if (!inDomain) {
     db.run('INSERT INTO domains (user_id, domain) VALUES(?, ?)', id, domain);
     await message.reply('You have been successfully added to ' + available_domains[domain]);
   } else {
--- a/src/components/interview.ts
+++ b/src/components/interview.ts
@@ -94,7 +94,7 @@ async function addInterviewer(message: Discord.Message, args: string[]): Promise
   const res = await db.get('SELECT * FROM interviewers WHERE user_id = ?', id);
   if (!res) {
     db.run('INSERT INTO interviewers (user_id, link) VALUES(? , ?)', id, parsedLink);
-    message.reply(`Your info has been added. Thanks for helping out! :codeyLove:`);
+    message.reply(`Your info has been added. Thanks for helping out!`);
   } else {
     db.run('UPDATE interviewers SET link = ? WHERE user_id = ?', parsedLink, id);
     message.reply(`Your info has been changed.`);
--- a/src/components/interview.ts
+++ b/src/components/interview.ts
@@ -14,6 +14,19 @@ interface Interviewer {
   link: string;
 }
 
+function domainString() {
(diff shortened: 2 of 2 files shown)


gzcharleszhang:
in order to link the issues I think you could use `resolves`, I don't think `completes` is a keyword github understands. `closes` also works i think

src/components/interview.ts
```diff
@@ -36,41 +52,72 @@ export async function handleInterview(message: Discord.Message, args: string[],
...
+    ' ```.interviewer domain [domain] \n//adds or removes you from a domain``` ';
   await message.channel.send(response);
 }
 
 async function addInterviewer(message: Discord.Message, args: string[]): Promise<void> {
   const db = await openDB();
   const id = message.author.id;
+
+  //checks for existence of link arg
   const link = args.shift();
```
gzcharleszhang, line 69: this takes the rightmost arg, should we maybe check the number of args supplied for better user input validation? again with Commando or slash commands we wouldn't have to worry about this.

src/components/interview.ts
```diff
@@ -36,41 +52,72 @@ export async function handleInterview(message: Discord.Message, args: string[],
...
     message.channel.send(`<@${id}>, your info has been changed.`);
   }
 }
 
-async function listInterviewers(message: Discord.Message, client: Discord.Client): Promise<void> {
+async function listInterviewers(message: Discord.Message, client: Discord.Client, args: string[]): Promise<void> {
   const db = await openDB();
-  const res = _.shuffle(await db.all('SELECT * FROM interviewers')) as Interviewer[];
+  let res: Interviewer[];
+  let domain = args.shift();
```
gzcharleszhang, line 96: same here, should we maybe validate the number of arguments supplied? if all these commands only require 1 or less arguments, maybe we could do a top level check somewhere for all of these subcommands

gzcharleszhang: maybe like `const domain = args.length == 1 ? args[0].toLowerCase() else null` if you are planning to check number of args here

src/components/interview.ts
```diff
@@ -36,41 +52,72 @@ export async function handleInterview(message: Discord.Message, args: string[],
...
+  if (domain) domain = domain.toLowerCase();
+  let title = 'Available interviewers';
+  console.log(domain);
```
gzcharleszhang, line 99: don't forget to remove

src/components/interview.ts
```diff
@@ -36,41 +52,72 @@ export async function handleInterview(message: Discord.Message, args: string[],
...
+  //checks if domain arg is passed and is valid, runs list without filters otherwise
+  if (!domain || !(domain in available_domains)) {
+    if (domain) {
```
gzcharleszhang, line 102: these nested if statements is kind of complex, maybe something like
```
if (!domain) {
  // query for all interviewers
} else if (!(domain in available_domains)) {
  // send error message for invalid domain
} else {
  // query for specific domains
}
```

src/components/interview.ts
```diff
@@ -80,3 +127,80 @@ async function listInterviewers(message: Discord.Message, client: Discord.Client
   outEmbed.setDescription(listString);
   await message.channel.send(outEmbed);
 }
+
+async function clearProfile(message: Discord.Message): Promise<void> {
+  const db = await openDB();
+  const id = message.author.id;
```
gzcharleszhang, line 133: `const { id } = message.author;`

src/components/interview.ts
```diff
@@ -80,3 +127,80 @@ async function listInterviewers(message: Discord.Message, client: Discord.Client
...
+
+  //clear user data from both tables
+  await db.run('DELETE FROM interviewers WHERE user_id = ?', id);
+  await db.run('DELETE FROM domains WHERE user_id = ?', id);
+  await message.channel.send('Your interviewer data has been cleared!');
+}
+
+async function showProfile(message: Discord.Message): Promise<void> {
+  const db = await openDB();
+  const id = message.author.id;
```
gzcharleszhang, line 143: `const { id } = message.author;`

src/components/interview.ts
```diff
@@ -80,3 +127,80 @@ async function listInterviewers(message: Discord.Message, client: Discord.Client
...
+  if (!interviewer) {
+    await message.channel.send("You don't seem to have signed up yet, run ```.interviewer signup [link]```");
+    return;
+  }
+
+  //get user's domains (if any)
+  const res = await db.all('SELECT * FROM domains WHERE user_id = ?', id);
+  let domains = '';
+  for (const i of res) {
+    domains += available_domains[i.domain] + ' ';
```
gzcharleszhang, line 156: maybe comma separated is more readable

src/components/interview.ts
```diff
@@ -80,3 +127,80 @@ async function listInterviewers(message: Discord.Message, client: Discord.Client
...
+  //build output embed
+  const outEmbed = new Discord.MessageEmbed().setColor('#0099ff').setTitle('Interviewer Profile');
+  outEmbed.addField('**Link**', interviewer['link']);
+  outEmbed.addField('**Domains**', domains === '' ? 'None' : domains);
+  await message.channel.send(outEmbed);
+}
+
+async function addDomain(message: Discord.Message, args: string[]): Promise<void> {
+  const db = await openDB();
+  const id = message.author.id;
```
gzcharleszhang, line 168: `const { id } = message.author`

src/components/interview.ts
```diff
@@ -80,3 +127,80 @@ async function listInterviewers(message: Discord.Message, client: Discord.Client
...
+
+  //get domain and check arg passed
+  let domain = args.shift();
```
gzcharleszhang, line 171: same here, looks like we always assume args only has 1 element, maybe we could validate this somewhere higher level.

src/components/interview.ts
```diff
@@ -80,3 +127,80 @@ async function listInterviewers(message: Discord.Message, client: Discord.Client
...
+  if (!domain) {
+    help(message);
+    return;
+  }
+
+  //parse domain to lowercase and check if valid
+  domain = domain.toLowerCase();
+  if (!(domain in available_domains)) {
+    let output = 'Not a valid domain, valid domains are:';
+    for (const i of Object.values(available_domains)) {
```
gzcharleszhang, line 181: looks like we built this available domains string a few times, maybe we could refactor this into its own method so we could reuse the same functionality

gzcharleszhang (changes requested):
making some big changes :eyes:

src/components/interview.ts
```diff
@@ -36,41 +52,72 @@ export async function handleInterview(message: Discord.Message, args: string[],
...
     message.channel.send(`Hmmm... I don't seem to recognize your meeting link. Be sure to use calendly or x.ai.`);
     return;
   }
+
+  //checks if user is already an interviewer, adds/updates info accordingly
   const res = await db.get('SELECT * FROM interviewers WHERE user_id = ?', id);
   if (!res) {
     db.run('INSERT INTO interviewers (user_id, link) VALUES(? , ?)', id, parsedLink);
-    message.channel.send(`<@${id}>, your info has been added. Thanks for signing up to help out! :codeyLove:`);
+    message.channel.send(`<@${id}>, your info has been added. Thanks for helping out! :codeyLove:`);
```
gzcharleszhang, line 86: might be worth looking into reply instead of manually tagging the author https://discord.js.org/#/docs/main/stable/class/Message?scrollTo=reply

src/components/interview.ts
```diff
@@ -36,41 +52,72 @@ export async function handleInterview(message: Discord.Message, args: string[],
 }
 
 async function help(message: Discord.Message): Promise<void> {
+  //should be deprecated as soon as proper help handler is built.
   const response =
     "I can't seem to recognize your command; the commands I know for interviewers are: \n" +
-    ' ```.interviewer signup [calendly/xai link] ``` \n' +
-    ' ```.interviewer list ```';
+    ' ```.interviewer signup [calendly/xai link] \n//signs up or update your calendar link```' +
```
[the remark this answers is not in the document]
alexz429: The amount of content in the code blocks makes it really questionable for inline code block. All else fails embeds might be a minor improvement.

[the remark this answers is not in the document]
gzcharleszhang: yeah I agree, if we use inline we probably need an embed. I just don't really like how code blocks look in the final response

src/components/db.ts
```diff
@@ -11,12 +11,13 @@ export const openDB = async (): Promise<Database> => {
...
     });
     //initialize all relevant tables
-    await db.run('CREATE TABLE IF NOT EXISTS saved_data (msg_id INTEGER PRIMARY KEY,data TEXT NOT NULL);');
+    await db.run('CREATE TABLE IF NOT EXISTS saved_data (msg_id INTEGER PRIMARY KEY, data TEXT NOT NULL);');
     await db.run(
       'CREATE TABLE IF NOT EXISTS suggestions (id INTEGER PRIMARY KEY NOT NULL, author VARCHAR(255) NOT NULL,' +
         'created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP, suggestion TEXT NOT NULL, state INTEGER NOT NULL);'
     );
     await db.run('CREATE TABLE IF NOT EXISTS interviewers (user_id TEXT PRIMARY KEY, link TEXT NOT NULL)');
+    await db.run('CREATE TABLE IF NOT EXISTS domains (user_id TEXT NOT NULL, domain TEXT NOT NULL)');
```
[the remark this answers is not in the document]
kallentu: lmao Alex is going to take ad-hoc CS 348

src/components/interview.ts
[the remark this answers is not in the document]
kallentu: Can we try with embeds and see what this looks like? It also might be worth putting screenshots in your PR description so we can see without testing (assuming the ss are updated)

src/components/interview.ts
```diff
@@ -80,3 +127,80 @@ async function listInterviewers(message: Discord.Message, client: Discord.Client
...
+    let output = 'Not a valid domain, valid domains are:';
+    for (const i of Object.values(available_domains)) {
+      output += ' ' + i;
+    }
+    await message.channel.send(output);
+    return;
+  }
+
+  //check user signed up to be an interviewer
+  const interviewer = await db.get('SELECT * FROM interviewers WHERE user_id = ?', id);
```
kallentu, line 189: Can we do this interviewer check before domain? Then we can do one less check if people aren't even signed up.

src/components/db.ts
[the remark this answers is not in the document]
alexz429: Added (hopefully I did it right) index to domains table for domain column, which should be the most searched.

alexz429 (commented):
Added table indices and switched message.channel.send with message.reply. Functionally, it does the same thing (direct message replies aren't supported in discord.js yet), but simplifies the codebase.

gzcharleszhang:
yeah I think the reply functionality might be introduced in v13 (with slash commands), but at least now we don't have to do that @ mention hack with user ID

src/components/interview.ts
```diff
@@ -3,14 +3,34 @@ import Discord from 'discord.js';
...
 interface Interviewer {
-  user_id: number;
+  user_id: string;
   link: string;
 }
 
+function domainString() {
+  let output = '';
+  let first_domain = true;
+  for (const i of Object.values(available_domains)) {
```
gzcharleszhang, line 20: maybe using `join` would help make this cleaner https://lodash.com/docs/4.17.15#join

src/components/interview.ts
```diff
@@ -3,14 +3,25 @@ import Discord from 'discord.js';
...
+  design: 'Design',
+  pm: 'PM'
+};
 interface Interviewer {
-  user_id: number;
+  user_id: string;
   link: string;
 }
 
+function domainString() {
```
gzcharleszhang, line 17: this function could take a `domains` parameter, so it's more reusable

gzcharleszhang: nit: rename to `getDomainsString`, a function should preferably start with a verb/action word

src/components/interview.ts
```diff
@@ -36,47 +56,157 @@ export async function handleInterview(message: Discord.Message, args: string[],
...
+  outEmbed.addField('**Domains**', domains === '' ? 'None' : domains);
+  await message.reply(outEmbed);
+}
+
+async function addDomain(message: Discord.Message, args: string[]): Promise<void> {
+  const db = await openDB();
+  const { id } = message.author;
+
+  //check user signed up to be an interviewer
+  const interviewer = await db.get('SELECT * FROM interviewers WHERE user_id = ?', id);
```
gzcharleszhang, line 182: this check/reply is used in multiple places, maybe this could be its own helper?

src/components/interview.ts
```diff
@@ -36,47 +56,157 @@ export async function handleInterview(message: Discord.Message, args: string[],
...
+  }
+
+  //check if domain valid
+  if (!(domain in available_domains)) {
+    await message.reply('Not a valid domain, valid domains are: ' + domainString());
+    return;
+  }
+
+  //check if user already in domain
+  const res = await db.get('SELECT * FROM domains WHERE user_id = ? AND domain = ?', id, domain);
```
gzcharleszhang, line 202: nit: rename to `inDomain` for better readability

gzcharleszhang (changes requested):
good work, just some comments to improve reusability and code style

src/components/interview.ts
gzcharleszhang: in this case arrow functions might look cleaner
```
const getDomainsString = domains => _.join(domains, ', ');
```
I kinda think we should use arrow functions for the other functions too just to be consistent.

src/components/interview.ts
alexz429: I'm not sure if that would make the code any simpler, since its already one line. I'll also have to overload it a ton since different sql commands have different variables (alternatively, if I did a format string, it would probably be more complex)

gzcharleszhang: If you're worried about the return type, you could just return a boolean and do the check inside the helper, some like `doesInterviewerExist`. Normally yes, this is already really simple, but generally it's good practice to not have the same SQL queries defined in multiple places. Does that make sense?

kallentu (approved):
LGTM, pending Charles' styling comments.

alexz429:
made new doesQueryExist method. Had plans on making db a parameter too so it will dynamically search whatever we want, but weird typescript errors of Database type does not really check out right now... Since we only really have one db I thought it should be sufficient for the next while.

alexz429:
changed the idea of doesInterviewerExists to getInterviewer, since some methods still require the interviewer object afterwards for logic. The sql query is still modularized so it still reduces the number of "magic strings" in the code.

gzcharleszhang:
@alexz429 yeah that's actually a better idea, good job. Let me take another look at the changes.

gzcharleszhang (approved):
LGTM

PR #22, opened 2021-06-17
Commando framework with suggest and mock interview commands

### LINKED ISSUE #11: Revamp current commands framework
Currently, we have a giant switch statement that handles the branching of command logic. This is not scalable and quite messy. We want to define a CommandHandler class and a CommandExecutor class, so a CommandExecutor instance can add an arbitrary number of CommandHandlers and create a mapping of command to handler functions.

We could also use discord.js's `Commando` framework.

gzcharleszhang:
# Related Issues
resolves #11 

# Summary of Changes 
- Using Commando client instead of Discord.js client.
- Commands are organized into groups, each group is a directory in `src/commands`, and each command will live in a group directory. The groups we added so far are `suggestions` and `interviews`.
- Owners of the client has access to admin commands such as disabling/enabling command groups.
- New commands can be created by defining a new class that inherit from `BaseCommand` or `AdminCommand` in `src/utils/commands.ts`.
- Sub commands are no longer space separated, e.g. `.interviewer signup` is now `.interviewer-signup`.
- Each sub command separated by a dash also has an alias without the dash e.g. `.interviewersignup`.

# Steps to Reproduce
- try `.suggest` and wait for message prompt
- try interviewer commands such as `.interviewer-signup`
- try `.help`
- try `.help interviewer-signup`

# Demo
## Global help page
image

## Missing argument prompt
image

## Command specific help page
image


src/commands/interviews/interviewers.ts
```diff
@@ -0,0 +1,61 @@
...
+          key: 'domain',
+          prompt: `enter one of ${getAvailableDomainsString()}.`,
+          type: 'string',
+          default: '',
+          validate: validateDomainArg,
+          parse: parseDomainArg
+        }
+      ],
+      description: 'Shows you a list of available interviewers.',
+      examples: ['.interviewer signup calendly.com/codey/mock-interview']
```
alexz429, line 30: Should be examples of interviewer list, not interviewer signup.

gzcharleszhang: good catch :)

src/components/interview.ts
```diff
@@ -35,177 +42,86 @@ function parseLink(link: string) {
...
-  await message.reply(response);
-}
+};
 
-async function addInterviewer(message: Discord.Message, args: string[]): Promise<void> {
+/*
+  If interviewer doesn't exist, save the interviewer and their calendar link.
+  Otherwise, update the interviewer's calendar link.
+*/
+export const upsertInterviewer = async (id: string, calendarUrl: string): Promise<void> => {
```
alexz429, line 51: is upsert a thing?

gzcharleszhang: I've seen that word being used before and it's basically what we're trying to do https://en.wiktionary.org/wiki/upsert#:~:text=upsert%20(plural%20upserts),updates%20them%20if%20they%20do.

gzcharleszhang: oh it's a thing in postgres as well lol https://wiki.postgresql.org/wiki/UPSERT

src/commands/interviews/interviewers.ts
```diff
@@ -0,0 +1,61 @@
...
+  private async getInterviewerDisplayInfo(interviewer: Interviewer) {
+    const userTag = (await this.client.users.fetch(interviewer['user_id'])).tag;
+    return '**' + userTag + '** | [Calendar](' + interviewer['link'] + ')\n\n';
+  }
+
+  async onRun(message: CommandoMessage, args: { domain: string }): Promise<Message> {
+    const { domain } = args;
+
+    // query interviewers
+    const interviewers = await getInterviewers(args.domain);
```
marko-polo-cheno, line 43: Are there any issues if the query gives nothing?

src/components/interview.ts
```diff
@@ -35,177 +42,86 @@ function parseLink(link: string) {
...
+  return res.map((row) => availableDomains[row.domain]);
+};
 
-async function addDomain(message: Discord.Message, args: string[]): Promise<void> {
+/*
+  If domain already exists, remove interivewer from the domain and return true.
+  If domain doesn't exist, add interviewer to the domain and return false.  
+  Throws an error if domain isn't provided or not valid.
+*/
+export const toggleDomain = async (id: string, domain: string): Promise<boolean> => {
```
marko-polo-cheno, line 108: I think we only care if inDomain carries a value, but should it be a boolean? Or is it like that on purpose to kinda map empty to 0 and non-empty to 1...

src/commands/interviews/interviewerDomain.ts
```diff
@@ -0,0 +1,48 @@
...
+      examples: ['.interviewer-domain frontend', '.interviewerdomain pm']
+    });
+  }
+
+  async onRun(message: CommandoMessage, args: { domain: string }): Promise<Message> {
+    const { domain } = args;
+    const { id } = message.author;
+
+    //check user signed up to be an interviewer
+    if (!(await getInterviewer(id))) {
```
marko-polo-cheno, line 32: Whenever u kinda do a null check, or '?' conditional, I think it'd be good to try to be consistent in all the files but im just being extra cause idk what egg to find lol

marko-polo-cheno (commented):
where egg where egg

src/commands/interviews/interviewers.ts
gzcharleszhang: I don't think so, I think I tested when the domain doesn't have any interviewers

src/components/interview.ts
gzcharleszhang: yeah this just casts to bool, we could do a `!!inDomain` if it makes it clearer

src/commands/interviews/interviewerDomain.ts
gzcharleszhang: do you mean if statement vs ternary statement? I used if statement here because we only care about the case where there is not interviewer with that ID

src/components/interview.ts
gzcharleszhang: oh it actually doesn't, I'll have to cast it explicitly, good catch :)

src/commands/interviews/interviewerDomain.ts
marko-polo-cheno: mb lemme clarify - here u have if `!func()`
usually I saw 
`bool isCool = func();`
`if isCool do smth`

gzcharleszhang: I think usually we do that if we need the `isCool` variable later on, in this case we don't need the interviewer

src/commands/interviews/interviewers.ts
```diff
@@ -0,0 +1,61 @@
...
+    // shuffles interviewers to load balance
+    const shuffledInterviewers = _.shuffle(interviewers);
+    // only show up to page limit
```
kallentu, line 46: Is this part going to be an issue for when we want to move from page 1 to page 2 (with like the new buttons or something? We don't really have page states of what we've already seen after getting the shuffled interviewers. Not sure if that's just intended basic feature rn (which is fine for now I guess, but we will need to think about shuffling and then paginating)

Also, let me know if you don't understand me LOL I don't think I'm explaining this well.

src/components/interview.ts
kallentu: Hella weird terminology LOL

src/components/interview.ts
```diff
@@ -35,177 +42,87 @@ function parseLink(link: string) {
...
-  outEmbed.addField('**Domains**', domains === '' ? 'None' : domains);
-  await message.reply(outEmbed);
-}
+  const res = await db.all('SELECT domain FROM domains WHERE user_id = ?', id);
+  return res.map((row) => availableDomains[row.domain]);
+};
 
-async function addDomain(message: Discord.Message, args: string[]): Promise<void> {
+/*
+  If domain already exists, remove interivewer from the domain and return true.
```
kallentu, line 104: Nit: Comment typo ):

kallentu (changes requested):

src/commands/interviews/interviewers.ts
gzcharleszhang: I thought we decided to not pagination? Pagination with shuffling is much more difficult and is definitely outside the scope of this PR. We should either choose shuffling or pagination for now.

src/commands/interviews/utils.ts
```diff
@@ -0,0 +1,12 @@
+import { availableDomains, getAvailableDomainsString } from '../../components/interview';
+
+export const validateDomainArg = (value: string): boolean | string => {
+  // validate if this is one of the available domains
+  if (value.toLowerCase() in availableDomains) return true;
+  return `you entered an invalid domain, please enter one of ${getAvailableDomainsString()}.`;
```
marko-polo-cheno, line 6: English police has entered the chat. "Nu", the popo says.

src/utils/embeds.ts
```diff
@@ -0,0 +1 @@
+export const EMBED_COLOUR = '#0099ff';
```
marko-polo-cheno, line 1: What colour is this tho :O 
Is it Codey-Fuzz-Blue ?

src/utils/commands.ts
```diff
@@ -0,0 +1,71 @@
...
+    });
+  }
+}
+
+const getMessageLogData = (message: CommandoMessage): { [key: string]: string } => {
+  return {
+    authorId: message.author.id,
+    authorUsername: message.author.username,
+    messageStr: message.toString(),
+    command: message.command?.name || '',
```
marko-polo-cheno, line 55: Should we have a placeholder that isnt empty string?

src/utils/commands.ts
```diff
@@ -0,0 +1,71 @@
...
+    onCommandRun(message);
+    return this.onRun(message, args, fromPattern, result);
+  }
+
+  onError(
+    err: Error,
+    message: CommandoMessage,
+    _args: unknown,
+    _fromPattern: boolean,
+    _result?: ArgumentCollectorResult | undefined
```
marko-polo-cheno, line 27: ngl i dont get this file - too many _?:_:?_:?_:?_:?_:?_: lol

gzcharleszhang: the underscore just tells Typescript that we aren't using these args in the function, the question mark just means that this arg might not exist. I think we might be able to remove the `| undefined` here, let me check

src/utils/commands.ts
gzcharleszhang: I don't think that's a good idea, we need a good sentinel value to differentiate when a command exists or not

src/utils/embeds.ts
gzcharleszhang: no idea I just used whatever @alexz429 used before

src/commands/interviews/utils.ts
gzcharleszhang: what's wrong with this, capitalizing the `you`?

marko-polo-cheno: lol and split to 2 sentences

gzcharleszhang: I didn't capitalize because the bot replies like "@fishy, you entered an invalid domain, ..."

src/utils/embeds.ts
alexz429: Its def not the right blue, might have to ask Design on that one.

src/utils/commands.ts
```diff
@@ -0,0 +1,71 @@
...
+    return handleCommandError(err, message);
+  }
+
+  abstract onRun(
+    message: CommandoMessage,
+    // eslint-disable-next-line @typescript-eslint/ban-types
+    args: object | string | string[],
+    fromPattern: boolean,
+    result?: ArgumentCollectorResult
+  ): Promise<Message | Message[] | null> | null;
```
alexz429, line 38: How does a return value of null (without the promise) actually do anything? I was under the impression that the wrapped promise<null> is sufficient.

gzcharleszhang: When you have an async function, the return type has to be wrapped in a promise. `null` is just for non-async `onRun` where you return nothing.

gzcharleszhang: Now I think about it, I'm gonna remove `null` anyways, since Commando's underlying implementation is async so we should do the same

src/utils/commands.ts
```diff
@@ -0,0 +1,71 @@
+import { Message, Permissions } from 'discord.js';
+import Commando, { CommandoMessage, CommandoClient, CommandInfo, ArgumentCollectorResult } from 'discord.js-commando';
+import logger, { logError } from '../components/logger';
+
+export abstract class BaseCommand extends Commando.Command {
+  constructor(client: CommandoClient, info: CommandInfo) {
+    super(client, info);
+  }
+
+  run(
```
marko-polo-cheno, line 10: Can you explain the difference between `run()` and `onRun()` ? 
Is `onRun` a just a function header?

marko-polo-cheno: Did Github just repeat my comment lol?

src/commands/interviews/interviewerProfile.ts
```diff
@@ -0,0 +1,42 @@
+import { Message, MessageEmbed } from 'discord.js';
+import { CommandoClient, CommandoMessage } from 'discord.js-commando';
+import _ from 'lodash';
+
+import { BaseCommand } from '../../utils/commands';
+import { getDomains, getDomainsString, getInterviewer } from '../../components/interview';
+import { EMBED_COLOUR } from '../../utils/embeds';
+
+class InterviewerClearCommand extends BaseCommand {
```
marko-polo-cheno, line 9: Clear vs Profile

gzcharleszhang: good catch

src/utils/commands.ts
gzcharleszhang: `run` is an abstract method in Commando's command class that we have to implement, which is what Commando will run when a command is triggered. `onRun` is an abstract method in our own `BaseCommand` class, where any commands we implement that inherit/extend from `BaseCommand` will have to implement. In `BaseCommand`'s `run` method, you can see that we first log the command, then call `onRun`, so that's what happens when a command is triggeered.

gzcharleszhang: In OOP, an abstract method is kinda like a header like you said, where it doesn't contain an implementation. Any subclasses that inherit that class will need to implement any abstract methods it has.

marko-polo-cheno: Ah I see I see - Thanks Charles!

marko-polo-cheno (approved):
Not sure how the userPerms actually behaves, but changes look fine to me :)

Issue #9, opened 2021-06-09
Add command to list pending suggestions


gzcharleszhang:
This should be an admin-only command.

This should output an embed of pending suggestions submitted by users. There could be some pagination feature so that the embed is not huge if there are a lot of pending suggestions.

e.g. `.suggest list` or `.suggest pending`

PR #30, opened 2021-06-27
Created Suggestions List Command

marko-polo-cheno:
# Related Issues
Resolves: https://github.com/uwcsc/codeybot/issues/9

# Summary of Changes 
Created suggestions-list command which shows commands in the suggestions table. Can also list by state.
At the moment, there are only 'created' suggestions.

# Steps to Reproduce
Try ``.suggestions``
Try ``.suggestions cReaTed``

image


marko-polo-cheno:
NOTE: needs to be `AdminCommand` - to do later

src/commands/suggestions/suggestionList.ts
```diff
@@ -0,0 +1,57 @@
...
+import { EMBED_COLOUR } from '../../utils/embeds';
+import { parseListArg, validateListArg } from './utils';
+
+const RESULTS_PER_PAGE = 6;
+
+class SuggestionsListCommand extends AdminCommand {
+  constructor(client: CommandoClient) {
+    super(client, {
+      name: 'suggestions-list',
+      aliases: ['suggestions', 'suggestion-list', 'suggestions-list'],
```
gzcharleszhang, line 15: we could add `suggest-list` too

src/commands/suggestions/suggestionList.ts
```diff
@@ -0,0 +1,57 @@
...
+      group: 'suggestions',
+      memberName: 'list',
+      args: [
+        {
+          key: 'list',
```
gzcharleszhang, line 20: I think `state` is a better name for this

src/commands/suggestions/utils.ts
```diff
@@ -0,0 +1,12 @@
+import { availableLists, getAvailableListsString } from '../../components/suggestions';
+
+export const validateListArg = (value: string): boolean | string => {
+  // validate if this is one of the available domains
+  if (value.toLowerCase() in availableLists) return true;
+  return `you entered an invalid list. Please enter one of ${getAvailableListsString()}.`;
```
gzcharleszhang, line 6: I think `state` makes more sense than `list`

src/commands/suggestions/utils.ts
```diff
@@ -0,0 +1,12 @@
+import { availableLists, getAvailableListsString } from '../../components/suggestions';
+
+export const validateListArg = (value: string): boolean | string => {
+  // validate if this is one of the available domains
```
gzcharleszhang, line 4: not domains

src/components/suggestions.ts
```diff
@@ -1,13 +1,27 @@
...
 import { openDB } from './db';
 
-// All states of suggestion records
-export enum SuggestionStates {
-  Created = 1,
-  Rejected,
-  Pending,
-  Accepted
+// maps from key to readable string
+export const availableLists: { [key: string]: string } = {
```
gzcharleszhang, line 6: we should call this `suggestionStates`

src/commands/suggestions/suggestionList.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  private async getSuggestionDisplayInfo(suggestion: Suggestion) {
+    return '**' + suggestion['id'] + '** | ' + suggestion['state'] + ' | ' + suggestion['suggestion'] + '\n\n';
+  }
+
+  async onRun(message: CommandoMessage, args: { list: string }): Promise<Message> {
+    const { list } = args;
+
+    // query suggestions
+    const suggestions = await getSuggestions(args.list.toLowerCase());
```
gzcharleszhang, line 41: don't think we need to do `toLowerCase` here because our parser does that for our arguments, still check to make sure though

gzcharleszhang: also since you already have `const { list } = args` you don't need to do `args.list` anymore, and I think we should call this `state` instead of `list`

gzcharleszhang (commented):
I think we should change the argument `list` to `state`, which is more consistent with our DB model and `list` can be confusing. Also can you include a screenshot of what the embed looks like?

marko-polo-cheno:
> I think we should change the argument `list` to `state`, which is more consistent with our DB model and `list` can be confusing. Also can you include a screenshot of what the embed looks like?

I think list is more appropriate because its like `ls` list show right? We can also see it as a list of suggestions - a list of just created, a complete list, etc.
`state` is what i want to use to update states

gzcharleszhang:
Also we already have an action word list, it’s very confusing to have something like “suggestion-list <list>”

marko-polo-cheno:
> In the future when you add state transition commands, are you going to call that argument list? Also this is what the command is doing, query suggestions in a certain state. We never had a notion of a “list” of suggestions, we’ve only had a notion of suggestion states, I don’t think it would make sense to create these “lists” now.

In my head it'll be like ``.suggestion-update 12 actionable`` where 12 is its id

gzcharleszhang:
Yeah `actionable` is the state you want it to be in right, so it would be called `state`.

src/components/suggestions.ts
```diff
@@ -42,3 +56,21 @@ export const addSuggestion = async (
     [authorId, authorUsername, suggestion, state]
   );
 };
+
+export const getSuggestions = async (list: string | null): Promise<Suggestion[]> => {
+  const db = await openDB();
+  let res: Suggestion[];
+
+  if (!list) {
```
kallentu, line 64: If I'm not mistaken, list here is the state they are querying right? I think you and Charles have been talking about it, but it might be good to clarify. 

`list` here isn't semantically correct either. `state` or something else would make sense since we're querying `suggestionStates` already.

gzcharleszhang: yeah you prob should clean up all the `list` stuff into `state`

marko-polo-cheno: oops i missed this file - thanks ppls :)

src/commands/suggestions/utils.ts
```diff
@@ -0,0 +1,11 @@
+import { suggestionStates, getAvailableListsString } from '../../components/suggestions';
+
+export const validateState = (value: string): boolean | string => {
+  // validate if this is one of the available states
+  if (value.toLowerCase() in suggestionStates) return true;
+  return `you entered an invalid state. Please enter one of ${getAvailableListsString()}.`;
+};
+
+export const parseListArg = (value: string): string => {
```
gzcharleszhang, line 9: should rename to `parseStateArg` :)

src/components/suggestions.ts
```diff
@@ -1,13 +1,27 @@
...
+  created: 'Created',
+  pending: 'Pending',
+  rejected: 'Rejected',
+  actionable: 'Actionable',
+  accepted: 'Accepted'
+};
+
+export const getListsString = (List: string[]): string => _.join(List, ', ');
+
+export const getAvailableListsString = (): string => getListsString(Object.values(suggestionStates));
```
gzcharleszhang, line 16: `getStatesString` and `getAvailableStatesString`

gzcharleszhang (approved):
LGTM after those few method renames

kallentu (approved):

Issue #24, opened 2021-06-19
emoji loader setup function to allow bot to use emojis when sending messages.


alexz429:
Every emote is in the form of name:id, where id is unique across DISCORD. The only way to find the id is by querying the client to find all emojis of a server. It would help if this data is preloaded when the bot starts to prevent unnecessary client calls later.

PR #31, opened 2021-06-30
add emote loading function

alexz429:
# Related Issues
Resolves https://github.com/uwcsc/codeybot/issues/24

# Summary of Changes 
Added emote.ts to support emoji loading, with emotelist being the dict for emotes
Updated interviewer signup to add :codeyLove: again

# Steps to Reproduce
NOTE: your bot must be in at least a server with the emojis. :codeyLove: was added to test server.
run .interviewersignup
image


src/components/emotes.ts
```diff
@@ -0,0 +1,8 @@
+import { Emoji } from 'discord.js';
+import { CommandoClient } from 'discord.js-commando';
+
+export const emoteList: { [key: string]: Emoji } = {};
+
+export const initEmotes = (client: CommandoClient): void => {
+  client.emojis.cache.forEach((emoji) => (emoteList[emoji.name] = emoji));
```
gzcharleszhang, line 7: what emojis are in this cache? is it all emojis in every server that the bot has access to?

src/bot.ts
```diff
@@ -36,6 +37,7 @@ export const startBot = async (): Promise<void> => {
     });
     const notif = (await client.channels.fetch(NOTIF_CHANNEL_ID)) as Discord.TextChannel;
     notif.send('Codey is up!');
+    initEmotes(client);
```
gzcharleszhang, line 40: we should prob call `initEmotes` before we send the 'Code is up' notif

gzcharleszhang (approved):
LGTM besides a small comment

src/components/emotes.ts
```diff
@@ -0,0 +1,8 @@
+import { Emoji } from 'discord.js';
+import { CommandoClient } from 'discord.js-commando';
+
+export const emoteList: { [key: string]: Emoji } = {};
```
gzcharleszhang, line 4: You don't have to make this change, but for future reference, sometimes you would want to keep this variable private and maybe write a `getEmoteByName(name)` function that simply returns `emoteList[name]`. This simply hides the implementation detail of the data structure we decided to use to store the emotes. This pattern is similar to our `getDB` function, and it's something nice to think about what you write more complicated code later on.

gzcharleszhang: We don't need to make this change here because `emoteList` is simple enough

src/commands/interviews/interviewerSignup.ts
```diff
@@ -37,7 +38,7 @@ class InterviewerSignupCommand extends BaseCommand {
 
     // Add or update interviewer info
     await upsertInterviewer(id, parsedUrl);
-    return message.reply(`your info has been updated. Thanks for helping out!`);
+    return message.reply(`your info has been updated. Thanks for helping out! ` + emoteList['codeyLove'].toString());
```
kallentu, line 41: I actually don't know JS standard lol, do yall use template literals? It's more concise with `${emoteList['codeyLove']}`. It doesn't matter too much with end of string case like ^, but if we wanted variables within a sentence, defs easier.

src/components/emotes.ts
kallentu, line 7: Can we also stick with one naming also? `Emoji` vs `Emote`. I think Discord uses emoji, so maybe we'd want to as well?

src/commands/interviews/interviewerSignup.ts
gzcharleszhang: yeah agreed, should use template string so you don't need the `+`

src/components/emotes.ts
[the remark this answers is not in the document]
gzcharleszhang: I think we wanna pre-load/cache emojis for faster search, the `find` in the discordjs API is probably just a linear search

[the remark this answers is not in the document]
kallentu: Yeah, turns out it is https://discord.js.org/#/docs/collection/master/class/Collection
Alright, makes sense so long as it's not just empty work.

src/components/emotes.ts
alexz429: yeah, this way I was thinking we could always give a backup emote server for codey if needed.

src/components/emotes.ts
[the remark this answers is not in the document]
alexz429: Hmmm.... wow maybe I'm bad at reading documentation.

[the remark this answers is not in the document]
gzcharleszhang: what you're doing is fine, Kallen was just making sure you are not doing empty work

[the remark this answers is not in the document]
kallentu: Oh @alexz429 I noticed in your recent commit, you ended up changing this. You don't need to alter this, it's all good so long as you'll use it later, I just wanted to make sure we were doing it for a reason.

src/components/emojis.ts
```diff
@@ -0,0 +1,12 @@
+import { Emoji } from 'discord.js';
+import { CommandoClient } from 'discord.js-commando';
+
+let client: CommandoClient;
+
+export const initEmotes = (cli: CommandoClient): void => {
+  client = cli;
+};
+
+export const getEmoteByName = (name: string): Emoji | undefined => {
```
kallentu, line 10: Oh and streamline the naming conventions of the functions in this file also! (-> `initEmojis`)

gzcharleszhang (approved):

Issue #6, opened 2021-06-09
Notify users/channels when someone makes an anonymous suggestion


gzcharleszhang:
We should create a cron (scheduled) job that scans the database for any newly created suggestions, when new ones are detected, the job should send these suggestions to a specific channel. This channel ID could be configured as an environment variable in `.env` or reuse the current system notif channel ID.

Useful library:
https://github.com/kelektiv/node-cron

PR #34, opened 2021-07-02
Cron suggestion state

marko-polo-cheno:
# Related Issues
Resolves https://github.com/uwcsc/codeybot/issues/6.

# Summary of Changes 
Created a scheduled command that runs once every minute to change 'Created' suggestions into 'Pending' suggestions.
Makes a notif in the same channel for now after at least one suggestion state was changed.

# Steps to Reproduce
.suggest 3 random things, and then:
image
Ensure your new suggestions are created in the same minute, and after a minute, a notification will appear, showing the new suggestions.
.suggestions again to make sure states have changed to pending
image

### DIFF
--- a/src/components/cron.ts
+++ b/src/components/cron.ts
@@ -2,13 +2,15 @@ import { updateSuggestionCron } from './suggestions';
 import { TextChannel } from 'discord.js';
 import { CommandoClient } from 'discord.js-commando';
 
+const NOTIF_CHANNEL_ID_OLD: string = process.env.NOTIF_CHANNEL_ID_OLD || '.';
+
 const CronJob = require('cron').CronJob;
 
 // Checks for new suggestions every 10 mins
 export const suggestionCronJob = new CronJob('0 */10 * * * *', async function (client: CommandoClient) {
   const isSuggestionsChanged = await updateSuggestionCron();
   if (isSuggestionsChanged) {
-    const messageChannel = client.channels.cache.get(`701335585272627203`);
+    const messageChannel = client.channels.cache.get(NOTIF_CHANNEL_ID_OLD);
     if (messageChannel == undefined) {
       throw 'Bad Channel';
     } else if (messageChannel.type === 'text') {
--- a/src/components/cron.ts
+++ b/src/components/cron.ts
@@ -1,0 +1,20 @@
+import { updateSuggestionCron } from './suggestions';
+import { TextChannel } from 'discord.js';
+import { CommandoClient } from 'discord.js-commando';
+
+const CronJob = require('cron').CronJob;
+
+// Checks for new suggestions every 10 mins
+export const suggestionCronJob = new CronJob('0 */10 * * * *', async function (client: CommandoClient) {
+  const isSuggestionsChanged = await updateSuggestionCron();
+  if (isSuggestionsChanged) {
+    const messageChannel = client.channels.cache.get(`701335585272627203`);
+    if (messageChannel == undefined) {
+      throw 'Bad Channel';
+    } else if (messageChannel.type === 'text') {
+      (messageChannel as TextChannel).send(`New suggestions exist.`);
+    } else {
+      throw 'Bad Channel Type';
+    }
+  }
+});
--- a/src/components/cron.ts
+++ b/src/components/cron.ts
@@ -1,13 +1,12 @@
 import { updateSuggestionCron } from './suggestions';
 import { TextChannel } from 'discord.js';
 import { CommandoClient } from 'discord.js-commando';
+import { CronJob } from 'cron';
 
 const NOTIF_CHANNEL_ID_OLD: string = process.env.NOTIF_CHANNEL_ID_OLD || '.';
 
-const CronJob = require('cron').CronJob;
-
 // Checks for new suggestions every min
-export const createSuggestionCron: any = (client: CommandoClient) =>
+export const createSuggestionCron = (client: CommandoClient) =>
   new CronJob('0 */1 * * * *', async function () {
     const isSuggestionsChanged = await updateSuggestionCron();
     if (isSuggestionsChanged) {
--- a/src/components/cron.ts
+++ b/src/components/cron.ts
@@ -7,7 +7,7 @@ const NOTIF_CHANNEL_ID_OLD: string = process.env.NOTIF_CHANNEL_ID_OLD || '.';
 const CronJob = require('cron').CronJob;
 
 // Checks for new suggestions every 10 mins
-export const suggestionCronJob = new CronJob('0 */10 * * * *', async function (client: CommandoClient) {
+export const suggestionCronJob = new CronJob('0 */1 * * * *', async function (client: CommandoClient) {
   const isSuggestionsChanged = await updateSuggestionCron();
   if (isSuggestionsChanged) {
     const messageChannel = client.channels.cache.get(NOTIF_CHANNEL_ID_OLD);
--- a/src/components/cron.ts
+++ b/src/components/cron.ts
@@ -6,17 +6,18 @@ const NOTIF_CHANNEL_ID_OLD: string = process.env.NOTIF_CHANNEL_ID_OLD || '.';
 
 const CronJob = require('cron').CronJob;
 
-// Checks for new suggestions every 10 mins
-export const suggestionCronJob = new CronJob('0 */1 * * * *', async function (client: CommandoClient) {
-  const isSuggestionsChanged = await updateSuggestionCron();
-  if (isSuggestionsChanged) {
-    const messageChannel = client.channels.cache.get(NOTIF_CHANNEL_ID_OLD);
-    if (messageChannel == undefined) {
-      throw 'Bad Channel';
-    } else if (messageChannel.type === 'text') {
-      (messageChannel as TextChannel).send(`New suggestions exist.`);
-    } else {
-      throw 'Bad Channel Type';
+// Checks for new suggestions every min
+export const createSuggestionCron = (client: CommandoClient) =>
+  new CronJob('0 */1 * * * *', async function () {
+    const isSuggestionsChanged = await updateSuggestionCron();
+    if (isSuggestionsChanged) {
+      const messageChannel = client.channels.cache.get(NOTIF_CHANNEL_ID_OLD);
+      if (messageChannel == undefined) {
+        throw 'Bad Channel';
+      } else if (messageChannel.type === 'text') {
+        (messageChannel as TextChannel).send(`New suggestions exist.`);
+      } else {
+        throw 'Bad Channel Type';
+      }
     }
-  }
-});
+  });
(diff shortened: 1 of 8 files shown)


marko-polo-cheno:
I need help fixing lint

src/components/cron.ts
```diff
@@ -0,0 +1,22 @@
...
+
+const NOTIF_CHANNEL_ID_OLD: string = process.env.NOTIF_CHANNEL_ID_OLD || '.';
+
+// Checks for new suggestions every min
+export const createSuggestionCron = (client: CommandoClient) =>
+  new CronJob('0 */1 * * * *', async function () {
+    const isSuggestionsChanged = await updateSuggestionCron();
+    if (isSuggestionsChanged) {
+      const messageChannel = client.channels.cache.get(NOTIF_CHANNEL_ID_OLD);
+      if (messageChannel == undefined) {
```
gzcharleszhang, line 14: we can just do a `if (!messageChannel)` here

src/components/cron.ts
```diff
@@ -0,0 +1,22 @@
...
+        throw 'Bad Channel';
+      } else if (messageChannel.type === 'text') {
+        (messageChannel as TextChannel).send(`New suggestions exist.`);
```
gzcharleszhang, line 17: We should send the new suggestions here instead of just sending a simple notification

src/components/cron.ts
```diff
@@ -0,0 +1,22 @@
...
+import { TextChannel } from 'discord.js';
+import { CommandoClient } from 'discord.js-commando';
+import { CronJob } from 'cron';
+
+const NOTIF_CHANNEL_ID_OLD: string = process.env.NOTIF_CHANNEL_ID_OLD || '.';
+
+// Checks for new suggestions every min
+export const createSuggestionCron = (client: CommandoClient): CronJob =>
+  new CronJob('0 */1 * * * *', async function () {
+    const isSuggestionsChanged = await updateSuggestionCron();
```
gzcharleszhang, line 11: instead of marking all the created suggestions as pending here, we should probably first query for all the created suggestions, then notify/show these suggestions to the channel in an embed, then call a function to update these suggestions as pending

src/components/suggestions.ts
```diff
@@ -57,6 +57,26 @@ export const addSuggestion = async (
   );
 };
 
+export const updateSuggestionCron = async (state = 'pending', oldState = 'created'): Promise<boolean> => {
```
gzcharleszhang, line 60: with the above comment, this can just be a simple `updateSuggestionState`, which could take in a list of suggestion IDs and the target state as parameters

gzcharleszhang: this function is nice because your next ticket could reuse this function

src/components/cron.ts
marko-polo-cheno: I think i need to account for undefined no matter what

gzcharleszhang: does `!messageChannel` not handle undefined?

src/components/suggestions.ts
```diff
@@ -57,6 +62,21 @@ export const addSuggestion = async (
   );
 };
 
+export const updateSuggestionCron = async (ids: number[], state = 'pending'): Promise<void> => {
```
gzcharleszhang, line 65: should rename to `updateSuggstionsState`, having cron in the name couples this function with the cron component, which is not what we want

src/components/cron.ts
```diff
@@ -0,0 +1,29 @@
...
+import { CronJob } from 'cron';
+import { EMBED_COLOUR } from '../utils/embeds';
+
+const MOD_CHANNEL_ID: string = process.env.MOD_CHANNEL_ID || '.';
+
+// Checks for new suggestions every min
+export const createSuggestionCron = (client: CommandoClient): CronJob =>
+  new CronJob('0 */1 * * * *', async function () {
+    const createdSuggestions = await getSuggestions('created');
+    const createdSuggestionIds = createdSuggestions.map((a) => a.id).map(Number);
```
gzcharleszhang, line 13: we can do `.map(a => Number(a.id))` here so we don't need the second map

src/components/cron.ts
```diff
@@ -0,0 +1,29 @@
...
+import { CommandoClient } from 'discord.js-commando';
+import { CronJob } from 'cron';
+import { EMBED_COLOUR } from '../utils/embeds';
+
+const MOD_CHANNEL_ID: string = process.env.MOD_CHANNEL_ID || '.';
+
+// Checks for new suggestions every min
+export const createSuggestionCron = (client: CommandoClient): CronJob =>
+  new CronJob('0 */1 * * * *', async function () {
+    const createdSuggestions = await getSuggestions('created');
```
gzcharleszhang, line 12: I don't really like having just string literals here for our states, maybe create an enum of states? Kinda like what we had before but we can use string enums this time.

```
enum SuggestionState {
  Created = 'created'
  Pending = 'pending'
  ...
}
```

src/components/cron.ts
```diff
@@ -0,0 +1,29 @@
...
+    const createdSuggestionIds = createdSuggestions.map((a) => a.id).map(Number);
+    if (createdSuggestionIds != undefined && createdSuggestionIds.length != 0) {
```
gzcharleszhang, line 14: you can use `_.isEmpty` from lodash here

src/components/cron.ts
```diff
@@ -0,0 +1,29 @@
...
+      const messageChannel = client.channels.cache.get(MOD_CHANNEL_ID);
+      if (messageChannel == undefined) {
+        throw 'Bad Channel from env file';
+      } else if (messageChannel.type === 'text') {
+        await updateSuggestionCron(createdSuggestionIds);
```
gzcharleszhang, line 19: we should update states after we sent the embed to be safe

src/components/suggestions.ts
```diff
@@ -39,6 +39,11 @@ export const initSuggestionsTable = async (db: Database): Promise<void> => {
   );
 };
 
+// Get id and suggestion from a list of Suggestions
+export const getSuggestionPrintout = async (suggestions: Suggestion[]): Promise<string> => {
+  return suggestions.map((suggestion) => '**' + suggestion['id'] + '** | ' + suggestion['suggestion'] + '\n').join('');
```
gzcharleszhang, line 44: if you just want a newline between each suggestion, you could just do `.join('\n')` and remove the newline in the map function

src/components/cron.ts
gzcharleszhang: you can rename the current `suggestionStates` object to `readableSuggestionStates`

src/components/cron.ts
```diff
@@ -0,0 +1,29 @@
...
+
+// Checks for new suggestions every min
+export const createSuggestionCron = (client: CommandoClient): CronJob =>
+  new CronJob('0 */1 * * * *', async function () {
+    const createdSuggestions = await getSuggestions('created');
+    const createdSuggestionIds = createdSuggestions.map((a) => a.id).map(Number);
+    if (createdSuggestionIds != undefined && createdSuggestionIds.length != 0) {
+      const messageChannel = client.channels.cache.get(MOD_CHANNEL_ID);
+      if (messageChannel == undefined) {
+        throw 'Bad Channel from env file';
```
gzcharleszhang, line 17: Should change to something like `Bad channel ID`, we shouldn't assume that the ID is from an env file

gzcharleszhang:
Logic looks good! Just some comments about styling and code organization

src/components/cron.ts
marko-polo-cheno: don't think it does

gzcharleszhang: that doesn't make sense `!undefined` returns `true` so it should work

gzcharleszhang: even if you need to check for undefined specifically, you should use strict equality `===`

src/components/suggestions.ts
```diff
@@ -3,17 +3,25 @@ import { Database } from 'sqlite';
...
-export const suggestionStates: { [key: string]: string } = {
+export const suggestionStatesReadable: { [key: string]: string } = {
   created: 'Created',
   pending: 'Pending',
   rejected: 'Rejected',
   actionable: 'Actionable',
   accepted: 'Accepted'
 };
 
+export enum suggestionStates {
```
gzcharleszhang, line 14: nit: we capitalize enum name and remain singular by convention, i.e. `SuggestionState`

marko-polo-cheno: o

src/components/cron.ts
marko-polo-cheno: because `!undefined` returns `true` i shouldn't use `!msgCh` because i can't send msgs to an undefined channel

gzcharleszhang: Doesn’t this if statement check if it’s a bad channel? So true is what you want? that’s the same as == undefined

marko-polo-cheno: Message channel can be `undefined` or a `Channel`. Mhm, true is what I want, but not by `!undefined`.

image
image

marko-polo-cheno: oh bruh ok i think i get what u mean - i thought u meant I dont need an else-if

gzcharleszhang: oh no LOL

src/components/suggestions.ts
```diff
@@ -57,19 +70,37 @@ export const addSuggestion = async (
...
+  const db = await openDB();
+  // Update created suggestions to pending suggestions in DB via ids
+  ids.map(async function (id) {
+    await db.run(
+      `
+      UPDATE suggestions
+      SET state = ?
+      WHERE id = ?;
+      `,
+      [suggestionStatesReadable[state], id]
```
gzcharleszhang, line 83: I missed this before, but why do we store readable strings in the database? The readable stuff should be for UI only, I thought we were always storing the lowercase key for state

marko-polo-cheno: I think its easier to just store the readable stuff in the DB - I dont see why we shouldn't.

gzcharleszhang: I don't think this is easier, when a command handler receives a state param, it's parsed into the lowercase key, which is an enum value. If the database has readable strings, then you have to parse that parameter into a readable string. This means you are parsing stuff twice. Also, it's very bad practice to couple UI things with database stuff. What if we want to change the readable string into something else? That means we would have to do a DB migration to update all the rows.

gzcharleszhang: what you are doing here is already more complicated, if we store the lowercase key, we only need to pass in `state`, here you are passing in `suggestionStatesReadable[state]`

gzcharleszhang: Let's say if we want to do some DB operation on a particular state, we could just pass in something like `SuggestionState.SomeState`, instead of `suggestionStatesReadable[SuggestionState.SomeState]`

gzcharleszhang (approved):

PR #35, opened 2021-07-03
Created dockerfile for production environment

### LINKED ISSUE #8: Docker specify environment variables at runtime
We should specify environment variables at runtime instead of at build time

### LINKED ISSUE #25: Create a dockerfile for production build
Currently, our dockerfile creates an image that includes dev dependencies and starts the application in a dev environment. We will need a separate dockerfile to build a production env image.

gzcharleszhang:
# Related Issues
resolves #25
resolves #8 

# Summary of Changes 
Added a dockerfile for production environment, which uses the `pm2` runtime to run the bot instead of `node`. The production image is also leaner than the dev image.

# Steps to Reproduce
Building production image: `docker build . -f Dockerfile.prod -t codey-prod`
Running production container: `docker run -d --env-file .env -v $PWD/logs:/usr/app/logs -v $PWD/db:/usr/app/db codey-prod`

### DIFF
--- a/Dockerfile.prod
+++ b/Dockerfile.prod
@@ -1,0 +1,19 @@
+# Stage 1: Typescript build app
+FROM node:16-alpine AS ts-build
+WORKDIR /usr/app
+COPY package.json .
+RUN npm install
+COPY . .
+RUN npm run ts:build
+
+# Stage 2: Run in production env
+FROM node:16-alpine
+WORKDIR /usr/app
+ENV NODE_ENV="prod"
+COPY package.json .
+RUN npm install --only=production
+RUN npm install pm2 -g
+COPY . .
+COPY --from=ts-build /usr/app/dist ./dist
+
+CMD [ "pm2-runtime", "--raw", "./dist/index.js" ]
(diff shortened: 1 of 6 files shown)


Dockerfile.prod
```diff
@@ -0,0 +1,19 @@
+# Stage 1: Typescript build app
+FROM node:16-alpine AS ts-build
+WORKDIR /usr/app
+COPY package.json .
+RUN npm install
+COPY . .
+RUN npm run ts:build
```
marko-polo-cheno, line 7: I dont entirely understand whats going on , but above u have `ts-build` and here u have `ts:build`
Other than that, LGTM

marko-polo-cheno (approved):

kallentu (approved):

Dockerfile.prod
gzcharleszhang: `ts:build` is an npm/yarn script we have in `package.json`, and `ts-build` above is just a label for this stage in the Dockerfile. We can pair and I can walk you through this if you want.

Issue #38, opened 2021-07-04
Add more interviewer domains


gzcharleszhang:
We should think of more interviewer domains and add them, like infra and data science

kallentu:
We can grab inspiration from Tech+ mentee/mentor sign up forms and CSC Project Program forms too.

gzcharleszhang:
Need more domains per offline discussion

PR #42, opened 2021-07-08
add domains and fix help messages

alexz429:
# Related Issues
Resolves https://github.com/uwcsc/codeybot/issues/38

# Summary of Changes 
Added new domains, and fixed domain help messages to be code blocks

# Steps to Reproduce
image

### DIFF
--- a/src/components/interview.ts
+++ b/src/components/interview.ts
@@ -9,8 +9,12 @@ export const availableDomains: { [key: string]: string } = {
   backend: 'Backend',
   design: 'Design',
   pm: 'PM',
-  datasci: 'Datasci',
-  infra: 'Infra'
+  datasci: 'DataSci',
+  infra: 'Infra',
+  mobile: 'Mobile',
+  research: 'Research',
+  hardware: 'Hardware',
+  gamedev: 'GameDev'
 };
 
 export interface Interviewer {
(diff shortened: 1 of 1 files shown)


src/components/interview.ts
```diff
@@ -29,9 +33,9 @@ export const getInterviewer = async (id: string): Promise<Interviewer | undefine
   return await db.get('SELECT * FROM interviewers WHERE user_id = ?', id);
 };
 
-export const getDomainsString = (domains: string[]): string => _.join(domains, ', ');
+export const getDomainsString = (domains: string[]): string => '`' + _.join(domains, '`, `') + '`';
```
gzcharleszhang, line 36: nit: use template string

alexz429: is it actually better in this case? I have to escape the backticks which make it look really awkward.

gzcharleszhang: I think in general we want to use template strings as much as possible. In this case the backticks made it uglier than normal but we should try to be consistent.

gzcharleszhang (approved):

src/components/interview.ts
```diff
@@ -9,8 +9,12 @@ export const availableDomains: { [key: string]: string } = {
   backend: 'Backend',
   design: 'Design',
   pm: 'PM',
-  datasci: 'Datasci',
-  infra: 'Infra'
+  datasci: 'DataSci',
+  infra: 'Infra',
```
kallentu, line 13: Stupid nit: Can we make this alphabetical? So it's easier to add/remove from and probably easier for the user to find their domain.

Issue #10, opened 2021-06-09
Add functionality to approve or reject pending suggestions


gzcharleszhang:
This can be done via reactions (e.g. check to accept, cross to reject) and/or via commands `suggestion accept <ID>` and `suggestion reject <ID>`

Useful resource for awaiting reactions:
https://discordjs.guide/popular-topics/collectors.html#message-collectors

marko-polo-cheno:
I'll keep it as text based for now

PR #46, opened 2021-07-15
Suggestion state update

marko-polo-cheno:
# Related Issues
https://github.com/uwcsc/codeybot/issues/10

# Summary of Changes 
Created Suggestion Update Command
.suggestions-update <state> <IDs>

# Steps to Reproduce
image
image

### DIFF
--- a/src/commands/suggestions/suggestionsUpdate.ts
+++ b/src/commands/suggestions/suggestionsUpdate.ts
@@ -11,8 +11,6 @@ import {
 import { EMBED_COLOUR } from '../../utils/embeds';
 import { parseStateArg, validateUpdate } from './utils';
 
-const RESULTS_PER_PAGE = 6;
-
 class SuggestionsUpdateCommand extends AdminCommand {
   constructor(client: CommandoClient) {
     super(client, {
--- a/src/commands/suggestions/suggestionsUpdate.ts
+++ b/src/commands/suggestions/suggestionsUpdate.ts
@@ -22,7 +22,7 @@ class SuggestionsUpdateCommand extends AdminCommand {
       memberName: 'update',
       args: [
         {
-          key: 'state',
+          key: 'value',
           prompt: `enter one of ${getAvailableStatesString()} before your IDs.`,
           type: 'string',
           default: '',
@@ -40,7 +40,7 @@ class SuggestionsUpdateCommand extends AdminCommand {
 
   async onRun(message: CommandoMessage, args: { value: string }): Promise<Message> {
     const { value } = args;
-    const values = value.split('\\s+');
+    const values = value.split(' ');
     const state = values[0] as SuggestionState;
     const ids = values.slice(1);
     const suggestionIds = ids.map((a) => Number(a));
--- a/src/commands/suggestions/suggestionsUpdate.ts
+++ b/src/commands/suggestions/suggestionsUpdate.ts
@@ -15,7 +15,7 @@ class SuggestionsUpdateCommand extends AdminCommand {
   constructor(client: CommandoClient) {
     super(client, {
       name: 'suggestions-update',
-      aliases: ['suggestions-update', 'suggestion-update'],
+      aliases: ['suggestion-update'],
       group: 'suggestions',
       memberName: 'update',
       args: [
@@ -23,7 +23,6 @@ class SuggestionsUpdateCommand extends AdminCommand {
           key: 'state',
           prompt: `enter one of ${getAvailableStatesString()}.`,
           type: 'string',
-          default: '',
           validate: validateState,
           parse: parseStateArg
         },
@@ -31,7 +30,6 @@ class SuggestionsUpdateCommand extends AdminCommand {
           key: 'ids',
           prompt: `enter suggestion IDs seperated by spaces.`,
           type: 'string',
-          default: '',
           validate: validateIDs
         }
       ],
@@ -45,9 +43,8 @@ class SuggestionsUpdateCommand extends AdminCommand {
 
   async onRun(message: CommandoMessage, args: { state: string; ids: string }): Promise<Message> {
     const { state, ids } = args;
-    const values = ids.split(' ');
+    const suggestionIds = ids.split(' ').map((a) => Number(a));
     const suggestionState = state as SuggestionState;
-    const suggestionIds = values.map((a) => Number(a));
 
     // Update states
     await updateSuggestionState(suggestionIds, suggestionState);
@@ -55,7 +52,7 @@ class SuggestionsUpdateCommand extends AdminCommand {
     // construct embed for display
     const title = `Suggestions Updated To ${suggestionStatesReadable[state]} State`;
     const outEmbed = new MessageEmbed().setColor(EMBED_COLOUR).setTitle(title);
-    outEmbed.setDescription(suggestionIds.join(' '));
+    outEmbed.setDescription(suggestionIds.join(', '));
     return message.channel.send(outEmbed);
   }
 }
--- a/src/commands/suggestions/suggestionsUpdate.ts
+++ b/src/commands/suggestions/suggestionsUpdate.ts
@@ -5,7 +5,8 @@ import { AdminCommand } from '../../utils/commands';
 import {
   suggestionStatesReadable,
   getAvailableStatesString,
-  updateSuggestionState
+  updateSuggestionState,
+  SuggestionState
 } from '../../components/suggestions';
 import { EMBED_COLOUR } from '../../utils/embeds';
 import { parseStateArg, validateUpdate } from './utils';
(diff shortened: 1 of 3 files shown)


src/commands/suggestions/suggestionsUpdate.ts
```diff
@@ -0,0 +1,57 @@
...
+  constructor(client: CommandoClient) {
+    super(client, {
+      name: 'suggestions-update',
+      aliases: ['suggestions-update', 'suggestion-update'],
+      group: 'suggestions',
+      memberName: 'update',
+      args: [
+        {
+          key: 'value',
+          prompt: `enter one of ${getAvailableStatesString()} before your IDs.`,
```
gzcharleszhang, line 24: this will take in the entire string though, for example `actionable 1 2 3`, is it possible to have an arg just for the state, and another arg for the IDs string?

src/commands/suggestions/suggestionsUpdate.ts
```diff
@@ -0,0 +1,57 @@
...
+      ]
+    });
+  }
+
+  async onRun(message: CommandoMessage, args: { value: string }): Promise<Message> {
+    const { value } = args;
+    const values = value.split(' ');
+    const state = values[0] as SuggestionState;
+    const ids = values.slice(1);
+    const suggestionIds = ids.map((a) => Number(a));
```
gzcharleszhang, line 44: nit: can combine the above two lines: `values.slice(1).map(...`

src/commands/suggestions/suggestionsUpdate.ts
```diff
@@ -36,20 +43,19 @@ class SuggestionsUpdateCommand extends AdminCommand {
...
-    const { value } = args;
-    const values = value.split(' ');
-    const state = values[0] as SuggestionState;
-    const ids = values.slice(1);
-    const suggestionIds = ids.map((a) => Number(a));
+  async onRun(message: CommandoMessage, args: { state: string; ids: string }): Promise<Message> {
+    const { state, ids } = args;
+    const values = ids.split(' ');
+    const suggestionState = state as SuggestionState;
+    const suggestionIds = values.map((a) => Number(a));
```
gzcharleszhang, line 50: you can just do `ids.split(' ').map(...` so you don't need `values`

src/commands/suggestions/utils.ts
```diff
@@ -1,16 +1,11 @@
...
-  const state = values[0];
-  // validate if first word is one of the available states
-  if (!(state.toLowerCase() in suggestionStatesReadable)) {
-    return `you entered an invalid state. Please enter one of ${getAvailableStatesString()}.`;
-  }
 
   // validate each id after first word
-  for (let i = 1; i < values.length; i++) {
+  for (let i = 0; i < values.length; i++) {
     if (isNaN(Number(values[i]))) {
```
gzcharleszhang, line 9: you can use lodash `some` instead of a loop here
```
if (_.some(values, isNaN)) {
  return `you entered an invalid id: ${values[i]}. Please enter numbers only.`;
}
```

src/commands/suggestions/suggestionsUpdate.ts
```diff
@@ -20,12 +20,19 @@ class SuggestionsUpdateCommand extends AdminCommand {
...
           default: '',
-          validate: validateUpdate,
+          validate: validateState,
           parse: parseStateArg
+        },
+        {
+          key: 'ids',
+          prompt: `enter suggestion IDs seperated by spaces.`,
+          type: 'string',
+          default: '',
```
gzcharleszhang, line 34: Since we have defaults for these two arguments, what happens if none is provided? If you think these arguments are required, you don't need default values for them.

gzcharleszhang: I'm curious what would happen if you just do `suggestionupdate pending` without specifying IDs

src/commands/suggestions/suggestionsUpdate.ts
```diff
@@ -0,0 +1,63 @@
...
+    const suggestionState = state as SuggestionState;
+    const suggestionIds = values.map((a) => Number(a));
+
+    // Update states
+    await updateSuggestionState(suggestionIds, suggestionState);
+
+    // construct embed for display
+    const title = `Suggestions Updated To ${suggestionStatesReadable[state]} State`;
+    const outEmbed = new MessageEmbed().setColor(EMBED_COLOUR).setTitle(title);
+    outEmbed.setDescription(suggestionIds.join(' '));
```
gzcharleszhang, line 58: I think adding a comma before the space might improve readability if there are a lot of numbers

src/commands/suggestions/utils.ts
gzcharleszhang: If you really want, this function can be a one liner
```
export const validateIDs = (value: string): boolean | string =>
  _.some(values.split(' '), isNaN) ? `you entered an invalid id: ${values[i]}. Please enter numbers only.` : true
```
Just make sure to add a comment about what it's doing.

kallentu (commented):
What does the help look like for suggestions in general?
I'm not fond of the current error message. Can we do a list like https://user-images.githubusercontent.com/39033120/121761214-7a752900-cafc-11eb-92d5-c4ef80802b77.png?

Also, what's the userflow like for writing `cancel`? do you literally just write cancel? What happens if you don't?

src/commands/suggestions/suggestionsUpdate.ts
marko-polo-cheno: ``suggestionupdate pending`` and ``suggestionupdate 1 2 3`` now give errors

marko-polo-cheno: image

marko-polo-cheno:
dmed .help visual
I think cancel is legit just typing cancel, otherwise after 30 seconds it'll automatically cancel. Beyond that, idk how it works

src/commands/suggestions/utils.ts
```diff
@@ -1,14 +1,13 @@
+import _ from 'lodash';
 import { suggestionStatesReadable, getAvailableStatesString } from '../../components/suggestions';
 
 export const validateIDs = (value: string): boolean | string => {
   // Split by white space
-  const values = value.split(' ');
+  const values = value.split(' ').map((a) => Number(a));
```
gzcharleszhang, line 6: why do we need to map each value to a number here? I thought `isNaN` below will just tell us if a string can't be parsed into a number?

gzcharleszhang (approved):
just one comment, everything else looks good

src/commands/suggestions/suggestionsUpdate.ts
gzcharleszhang: > suggestionupdate pending and suggestionupdate 1 2 3 now give errors

Do they give error or do they prompt for user input?

src/commands/suggestions/suggestionsUpdate.ts
```diff
@@ -0,0 +1,60 @@
...
+    });
+  }
+
+  async onRun(message: CommandoMessage, args: { state: string; ids: string }): Promise<Message> {
+    const { state, ids } = args;
+    const suggestionIds = ids.split(' ').map((a) => Number(a));
+    const suggestionState = state as SuggestionState;
+
+    // Update states
+    await updateSuggestionState(suggestionIds, suggestionState);
```
kallentu, line 50: Now that you're using this for general updates, please update/remove the comment in the function itself in `components/suggestions.ts` to reflect that it does more than just update created to pending.

kallentu: Also, are we just letting the ids that don't exist just fall through? We aren't planning on providing any warnings/errors for that?

src/commands/suggestions/suggestionsUpdate.ts
marko-polo-cheno: They error and give the cancel prompt (but at that point it'd be easier to just cancel)

src/commands/suggestions/utils.ts
marko-polo-cheno: image
Might be a typescript thing

src/commands/suggestions/suggestionsUpdate.ts
marko-polo-cheno: Mistyping a number (that exists) and messing up data is probs worse than changing ids that don't exist (does nothing to DB). Like there are no warnings if I turn some actually rejected suggestions back into created.
If we shouldn't have data mess-ups, we shouldn't have ids that DNE, so mods gotta be a bit more careful.

marko-polo-cheno: hmm if mistyping is easier than typing a num that doesnt exist...
@gzcharleszhang how might a ``confirm`` similar to how a ``cancel`` cancels work? (if its too complicated then lol just have good mods)

gzcharleszhang: let's just leave it like this for now

gzcharleszhang (changes requested):
Screen Shot 2021-07-20 at 10 59 57 AM
This is not good, we need to address this.

gzcharleszhang (approved):

PR #61, opened 2021-09-01
Bootcamp: Drop-in resume review event

asdera:
# Related Issues
Issue(s) that this PR will resolve, e.g. "resolves <issue link here>".

# Summary of Changes 
A summary or description of changes made in this PR.

# Steps to Reproduce
List the steps needed for the reviewer to produce your PR changes.


asdera:
image
fun...

asdera:
image
Important: This flag on codeybot must be ticked to allow it to listen for certain events

asdera:
Note: keyv package only works using require, disabled eslint rule temporalily

.eslintrc
```diff
@@ -7,6 +7,7 @@
     "plugin:prettier/recommended"
   ],
   "rules": {
-    "@typescript-eslint/no-unused-vars": ["warn", { "argsIgnorePattern": "^_" }]
+    "@typescript-eslint/no-unused-vars": ["warn", { "argsIgnorePattern": "^_" }],
+    "@typescript-eslint/no-var-requires": 0
```
gzcharleszhang, line 11: we don't need this once we remove keyv right?

src/bot.ts
```diff
@@ -38,13 +45,40 @@ export const startBot = async (): Promise<void> => {
...
       event: 'init'
     });
-    const notif = (await client.channels.fetch(NOTIF_CHANNEL_ID)) as Discord.TextChannel;
+
+    const notif = (await client.channels.fetch(NOTIF_CHANNEL_ID)) as TextChannel;
+
     initEmojis(client);
     createSuggestionCron(client).start();
+
+    // Intialize Bootcamp features if the event server exists
```
gzcharleszhang, line 54: could we put this section in its own function?

.eslintrc
asdera: yeah we won't need no-var-requires anymore, but the other rule was already there

src/bot.ts
asdera: yeah, i think we could just move all that logic into initBootcamp

src/utils/commands.ts
```diff
@@ -47,6 +47,15 @@ export abstract class AdminCommand extends BaseCommand {
   }
 }
 
+export abstract class MentorCommand extends BaseCommand {
```
gzcharleszhang, line 50: This seems specific to the bootcamp event, maybe move it to a bootcamp specific file instead? Also could you add some comments explaining who's considered a mentor and how that correlates to the perms required to use the command?

src/commands/bootcamp/bootcampKickMentee.ts
```diff
@@ -0,0 +1,48 @@
...
+      aliases: ['remove', 'boot'],
+      group: 'bootcamp',
+      memberName: 'kick',
+      guildOnly: true,
+      description: 'Kick mentees out of your 1 on 1 call.'
+    });
+  }
+
+  async onRun(message: CommandoMessage): Promise<Message> {
+    const mentor = message.member;
```
gzcharleszhang, line 18: cleaner to do `const { member: mentor, guild } = message`

src/commands/bootcamp/bootcampKickMentee.ts
```diff
@@ -0,0 +1,48 @@
...
+    const callChannel = mentor?.voice?.channel;
+    const guild = message.guild;
+    const track = callChannel?.parent?.name;
+
+    const queueVoice = <VoiceChannel>(
+      guild.channels.cache.find((channel) => channel.name === track && channel.type === 'voice')
```
gzcharleszhang, line 24: there should be a constant for `voice` in the discord.js lib, we should use taht instead

src/commands/bootcamp/bootcampKickMentee.ts
```diff
@@ -0,0 +1,48 @@
...
+      guildOnly: true,
+      description: 'Kick mentees out of your 1 on 1 call.'
+    });
+  }
+
+  async onRun(message: CommandoMessage): Promise<Message> {
+    const mentor = message.member;
+    const callChannel = mentor?.voice?.channel;
+    const guild = message.guild;
+    const track = callChannel?.parent?.name;
```
gzcharleszhang, line 21: why does parent mean here?

src/commands/bootcamp/bootcampKickMentee.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    const queueVoice = <VoiceChannel>(
+      guild.channels.cache.find((channel) => channel.name === track && channel.type === 'voice')
+    );
+    const queueChannel = <TextChannel>(
+      callChannel?.parent?.children.find(
```
gzcharleszhang, line 27: what is this doing here?

src/commands/bootcamp/bootcampKickMentee.ts
```diff
@@ -0,0 +1,48 @@
...
+        (channel) => channel.name === track?.replace(/ +/g, '-').toLocaleLowerCase() + '-queue'
+      )
+    );
+    const callMembers = callChannel?.members;
+
+    if (queueVoice && queueChannel && callMembers) {
+      if (callMembers?.size < 2) {
```
gzcharleszhang, line 34: magic number, make this a constant

src/commands/bootcamp/utils.ts
```diff
@@ -0,0 +1,16 @@
+import { max } from 'lodash';
+
+export const toTitleCase = (str: string): string => {
```
gzcharleszhang, line 3: does https://lodash.com/docs/4.17.15#startCase work?

src/commands/bootcamp/utils.ts
```diff
@@ -0,0 +1,16 @@
...
+  return str.replace(/-/g, ' ').replace(/\w\S*/g, function (txt: string) {
+    return txt.charAt(0).toUpperCase() + txt.substr(1).toLowerCase();
+  });
+};
+
+export const lowestInt = (nums: number[]): number => {
```
gzcharleszhang, line 9: is this the smallest int that's missing from `nums`? the function name could be better and documentation is needed

src/commands/bootcamp/bootcampNewTrack.ts
```diff
@@ -0,0 +1,80 @@
...
+      guildOnly: true,
+      description: 'Adds a new track category.'
+    });
+  }
+
+  async onRun(message: CommandoMessage, args: { category: string }): Promise<Message> {
+    let { category } = args;
+
+    category = toTitleCase(category);
+    const guild = message.guild;
```
gzcharleszhang, line 29: `const { guild } = message`

src/commands/bootcamp/bootcampNewTrack.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    const waitingRooms = <CategoryChannel>(
+      guild.channels.cache.find((channel) => channel.name.startsWith('Waiting Room') && channel.type === 'category')
```
gzcharleszhang, line 32: using magic strings like `Waiting Room` and `category` is kinda hacky, is there a discord.js constant for `category`?

src/commands/bootcamp/bootcampNewTrack.ts
```diff
@@ -0,0 +1,80 @@
...
+    );
+
+    if (!waitingRooms) {
+      guild.channels.create('Waiting Room', {
```
gzcharleszhang, line 36: should have a constant for `Waiting Room` since you are using this string literal more than once.

src/commands/bootcamp/bootcampNewTrack.ts
```diff
@@ -0,0 +1,80 @@
...
+class BootcampNewTrackCommand extends AdminCommand {
+  constructor(client: CommandoClient) {
+    super(client, {
+      name: 'new-track',
+      aliases: ['add-track'],
+      group: 'bootcamp',
+      memberName: 'new-track',
+      args: [
+        {
+          key: 'category',
```
gzcharleszhang, line 15: shouldn't this be `track` for consistency?

src/commands/bootcamp/bootcampNewTrack.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    const trackCategory = <CategoryChannel>(
+      guild.channels.cache.find((channel) => channel.name === category && channel.type === 'category')
+    );
+
+    if (trackCategory) {
+      return message.say('This channel category already exists.');
+    } else {
+      guild.channels
```
gzcharleszhang, line 49: need comment describing what this is doing here

src/commands/bootcamp/bootcampNewTrack.ts
```diff
@@ -0,0 +1,80 @@
...
+          type: 'voice',
+          parent: waitingRooms
+        })
+        .catch(console.error);
+
+      guild.channels
+        .create(category, {
+          type: 'category'
+        })
+        .then((newCategory) => {
```
gzcharleszhang, line 60: can we use `await` instead of `.then`?

src/commands/bootcamp/bootcampNewTrack.ts
```diff
@@ -0,0 +1,80 @@
...
+          guild.channels
+            .create(category?.replace(/ +/g, '-').toLocaleLowerCase() + '-queue', {
```
gzcharleszhang, line 62: I've seen this logic before, can we make this string manipulation magic a helper function to abstract away the complexity here?

src/commands/bootcamp/bootcampNewTrack.ts
```diff
@@ -0,0 +1,80 @@
...
+              parent: newCategory,
+              permissionOverwrites: [
+                {
+                  id: guild.roles.everyone,
+                  deny: [Permissions.FLAGS.VIEW_CHANNEL]
```
gzcharleszhang, line 67: what does this achieve? need to document

src/commands/bootcamp/bootcampNewTrack.ts
```diff
@@ -0,0 +1,80 @@
...
+                }
+              ]
+            })
+            .catch(console.error);
```
gzcharleszhang, line 71: please use logger to log error, this is not enough

gzcharleszhang: or you can just let the error get propagated upwards and let our default command error handling take care of it

src/commands/bootcamp/bootcampNextMentee.ts
```diff
@@ -0,0 +1,79 @@
...
+      aliases: ['mentee'],
+      group: 'bootcamp',
+      memberName: 'next',
+      guildOnly: true,
+      description: 'Moves the next mentee in line to your call.'
+    });
+  }
+
+  async onRun(message: CommandoMessage): Promise<Message> {
+    const mentor = message.member?.voice;
```
gzcharleszhang, line 19: why is `mentor` a voice state?

src/commands/bootcamp/bootcampNextMentee.ts
```diff
@@ -0,0 +1,79 @@
...
+    const callChannel = mentor?.channel;
+    const guild = message.guild;
```
gzcharleszhang, line 21: `const { guild } = message`, try to destructure as much as you can

src/commands/bootcamp/bootcampNextMentee.ts
```diff
@@ -0,0 +1,79 @@
...
+    const track = callChannel?.parent?.name;
+
+    let front;
```
gzcharleszhang, line 24: default value?

src/commands/bootcamp/bootcampNextMentee.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    const queueVoice = <VoiceChannel>(
+      guild.channels.cache.find((channel) => channel.name === track && channel.type === 'voice')
+    );
+    const queueChannel = <TextChannel>(
```
gzcharleszhang, line 29: I've seen this logic before, can we make these two helper functions?

src/commands/bootcamp/bootcampNextMentee.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    if (queueVoice && queueChannel) {
+      if (callChannel?.full) {
+        return message.say('You should only have one Mentee at a time.');
+      }
+      if (queueMembers.size < 1) {
+        return message.say('The are no Mentees waiting for this track.');
+      }
+
+      queueChannel.messages.fetch({ limit: queueVoice.members.size * 2 + 5 }).then((messages) => {
```
gzcharleszhang, line 44: this block of code definitely needs documentation

gzcharleszhang, line 44: `queueVoice.members.size * 2 + 5` abstract into its own variable/constant

gzcharleszhang, line 44: `await` instead of `.then`

src/commands/bootcamp/bootcampNextMentee.ts
```diff
@@ -0,0 +1,79 @@
...
+              const chatChannel = <TextChannel>(
+                mentorCall?.parent?.children.find(
+                  (channel) => channel.name === mentorCall?.name.replace(/ +/g, '-').toLocaleLowerCase() + '-vc'
+                )
+              );
+              chatChannel
+                .updateOverwrite(front, {
+                  VIEW_CHANNEL: true
+                })
+                .then(async () => {
```
gzcharleszhang, line 62: use await instead of .then

src/commands/bootcamp/bootcampNextMentee.ts
```diff
@@ -0,0 +1,79 @@
...
+                  const critiqueTime = (await BootcampSettings?.get('critique_time')) || 30;
```
gzcharleszhang, line 63: what is the 30 for? use constants please

src/commands/bootcamp/bootcampKickMentee.ts
```diff
@@ -0,0 +1,48 @@
...
+        return message.say('You are already alone in this call.');
+      }
+      callMembers
+        ?.filter((member) => member.id != mentor?.id)
+        .forEach((member) => {
+          member.voice.setChannel(null);
```
gzcharleszhang, line 40: do we need to set their view permissions of the text channel, can they see the channel after they're kicked?

gzcharleszhang (changes requested):
1. Your changes barely have any documentation/comments, this makes the code super hard to read and understand, please add more documentation and comments. Please also document any assumptions and business logic behind the code you write, for example, the assumption around naming conventions of voice/text channels.
2. Try to avoid using string literals and magic numbers as much as you can, either make them into constants or fully document what the numbers are (UI strings are an exception since we don't care about i18n).
3. Try to abstract repeated code into helper functions, repeated logic is really hard to refactor and is prone to error.
4. We prefer to use async-await instead of .then, so make sure this is consistent across our codebase.
5. Use [object/array destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) to make your code cleaner.
6. You should always think about how to achieve the same result by writing as few lines of code as possible.

gzcharleszhang:
Please also complete the PR description before requesting for review

src/commands/bootcamp/bootcampNextMentee.ts
```diff
@@ -0,0 +1,79 @@
...
+      if (callChannel?.full) {
+        return message.say('You should only have one Mentee at a time.');
+      }
+      if (queueMembers.size < 1) {
+        return message.say('The are no Mentees waiting for this track.');
+      }
+
+      queueChannel.messages.fetch({ limit: queueVoice.members.size * 2 + 5 }).then((messages) => {
+        messages
+          .sorted((mesgA, mesgB) => mesgA.createdTimestamp - mesgB.createdTimestamp)
```
gzcharleszhang, line 46: I can see this running into performance issues with large enough numbers, but a fix would be non trivial given our timelines

src/commands/bootcamp/bootcampNextMentee.ts
```diff
@@ -0,0 +1,79 @@
...
+          .every((msg): boolean => {
```
gzcharleszhang, line 47: why do you need to do a `every` when all you want is the `front`?

PR #76, opened 2021-09-26
change deploy dir to absolute path

### LINKED ISSUE #77: Enable OS Login on GCP and move codey files outside of user home directory
Currently, we don't have OS Login enabled on GCP, which means users added to the codey-bot project won't be able to ssh into the VM instance. Our codey related files are in `/home/exec/codey`, which is tied to the `exec` user, we should move it to `/codey` instead.

gzcharleszhang:
# Related Issues
resolves #77 

# Summary of Changes 
- Changed the codey deployment path to an absolute path on the GCP machine.
- Instead of scp `docker-compose.yml` onto the VM instance, we now `curl` the file directly from GitHub instead.

# Steps to Reproduce
- Merge this PR into master.


gzcharleszhang:
Before, the bot files live in `/home/exec/codey` in the GCP VM, this is the home dir for the user `exec`. After enabling OS login, the username for the exec account changed to `exec_csclub_uwaterloo_ca`, so it doesn't make sense for those files to be under `/home/exec` anymore. Nonetheless, those files shouldn't be in any user's home dir either way, so I moved them to `/codey`, which is why I made the chances in this PR to update our CD jobs.

kallentu (approved):

PR #100, opened 2021-10-24
87 daily and minute bonus

### LINKED ISSUE #86: Implement activity bonus

marko-polo-cheno:
# Related Issues
Resolves <https://github.com/uwcsc/codeybot/issues/87>
resolves #86 

# Summary of Changes 
Added listener to all commands to see if a user is eligible for a daily/minute Codey coin bonus.

# Steps to Reproduce
.cc counts as activity.
Keep .cc @self for over a minute, and watch the coin count increase. 
image


src/components/coin.ts
```diff
@@ -72,3 +91,82 @@ export const adjustCoinBalanceByUserId = async (userId: string, amount: number):
...
+    bonusTypeInTable
+  );
+};
+
+/*
+  Get the time of the latest bonus applied to a user based on type
+*/
+export const latestBonusByUserId = async (userId: string, type: BonusType): Promise<UserCoinBonus[] | undefined> => {
+  const db = await openDB();
+  const res: UserCoinBonus[] | undefined = await db.get(
```
gzcharleszhang, line 117: if you use `get` I think it just returns a single result, so it wouldn't have type `UserCoinBonus[]`

src/components/coin.ts
```diff
@@ -72,3 +91,82 @@ export const adjustCoinBalanceByUserId = async (userId: string, amount: number):
...
+  const lastBonusOccurence = await latestBonusByUserId(userId, bonusOfInterest.type);
+  console.log(lastBonusOccurence);
+  if (lastBonusOccurence === undefined) {
+    console.log('140');
+    await adjustCoinBalanceByUserId(userId, bonusOfInterest.amount);
+    await updateUserBonusTableByUserId(userId, bonusType);
+    return true;
+  }
+
+  const lastBonusOccurenceTime = lastBonusOccurence[0]['last_granted'].getTime();
```
gzcharleszhang, line 146: follow-up from previous comment, no need for `[0]` anymore

src/components/coin.ts
```diff
@@ -72,3 +91,76 @@ export const adjustCoinBalanceByUserId = async (userId: string, amount: number):
...
+  return res;
+};
+
+/*
+  Time bonus
+*/
+export const timeBonusByUserId = async (userId: string, bonusType: BonusType): Promise<boolean> => {
+  const bonusOfInterest = coinBonusMap.get(bonusType);
+
+  if (bonusOfInterest === undefined) {
```
gzcharleszhang, line 130: Can we just do `!bonusOfInterest`, also why do we return true here? Can you add more comments in this functions please?

src/components/coin.ts
```diff
@@ -72,3 +91,76 @@ export const adjustCoinBalanceByUserId = async (userId: string, amount: number):
...
+    return true;
+  }
+
+  const lastBonusOccurence = await latestBonusByUserId(userId, bonusOfInterest.type);
+  if (lastBonusOccurence === undefined) {
+    await adjustCoinBalanceByUserId(userId, bonusOfInterest.amount);
```
gzcharleszhang, line 136: can you put a `TODO` here to wrap these two operations in a transaction? We don't want to do adjustment if the updateUserBonus op fails.

src/components/coin.ts
```diff
@@ -72,3 +91,76 @@ export const adjustCoinBalanceByUserId = async (userId: string, amount: number):
...
+    await updateUserBonusTableByUserId(userId, bonusType);
+    return true;
+  }
+
+  const lastBonusOccurenceTime = new Date(lastBonusOccurence['last_granted']).getTime();
+  const nowTime = new Date().getTime();
+  const cooldown = nowTime - bonusOfInterest.cooldown;
+
+  if (lastBonusOccurenceTime < cooldown) {
```
gzcharleszhang, line 145: this block is has the exact same logic as the if statement before, can we just combine these two if statements into 1?

src/components/messageListener.ts
```diff
@@ -0,0 +1,8 @@
+import { Message } from 'discord.js';
+import { applyBonusByUserId } from './coin';
+
+export const messageListener = async (message: Message): Promise<void> => {
+  if (message.author.id != '843696481826373633') {
```
gzcharleszhang, line 5: hard coded ID?

gzcharleszhang: we may also want to make sure the author is not a bot

marko-polo-cheno: wups I forgot to change this
added to .env
also changed readme

Picowchew: In addition, I think it would be good if only guild messages lead to a user gaining Codey coins, since currently, a user can also gain Codey coins via DMs with the bot.

gzcharleszhang: I'm pretty sure you can do `client.user.id` to get the ID. Let's try to avoid env vars as much as possible because it involves configuration values for each deployment environment.

gzcharleszhang: https://discord.js.org/#/docs/main/v12/class/Client?scrollTo=user

src/components/coin.ts
```diff
@@ -72,3 +91,75 @@ export const adjustCoinBalanceByUserId = async (userId: string, amount: number):
...
+  return res;
+};
+
+/*
+  Adjusts balance by bonus amount if appropriate cooldown has passed.
+  Returns bool. 
+  True initiates the break in applyBonusByUserId, preventing multiple bonuses for a single msg.
+  False means this bonusType is not applicable yet.
+*/
+export const timeBonusByUserId = async (userId: string, bonusType: BonusType): Promise<boolean> => {
```
gzcharleszhang, line 130: maybe rename to `applyTimeBonus`?

src/components/coin.ts
```diff
@@ -72,3 +91,75 @@ export const adjustCoinBalanceByUserId = async (userId: string, amount: number):
...
+    userId,
+    type
+  );
+  return res;
+};
+
+/*
+  Adjusts balance by bonus amount if appropriate cooldown has passed.
+  Returns bool. 
+  True initiates the break in applyBonusByUserId, preventing multiple bonuses for a single msg.
```
gzcharleszhang, line 127: IMO this is breaking the separation of concern with `applyBonusByUserId`, this function shouldn't worry about how the other function will use the value of this function.

gzcharleszhang (approved):

.vscode/settings.json
```diff
@@ -3,5 +3,6 @@
   "editor.formatOnSave": true,
   "[javascript]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
-  }
+  },
+  "files.eol": "\n"
```
gzcharleszhang, line 7: 🧠

PR #109, opened 2021-11-06
Mention interviewers in interviewer list

### LINKED ISSUE #108: Mention interviewers in interviewer list
Currently, only the user tag is displayed. Mentioning would be more convenient.

Picowchew:
# Related Issues
Resolves #108 

# Summary of Changes 
Mention the interviewers in the interviewer list instead of displaying their user tags

# Steps to Reproduce
Run `.int`

### DIFF
--- a/src/commands/interviews/interviewers.ts
+++ b/src/commands/interviews/interviewers.ts
@@ -40,7 +40,8 @@ class InterviewersCommand extends BaseCommand {
   }
 
   private async getInterviewerDisplayInfo(interviewer: Interviewer) {
-    return `${await this.client.users.fetch(interviewer['user_id'])} | [Calendar](${interviewer['link']})\n\n`;
+    const user = await this.client.users.fetch(interviewer['user_id']);
+    return `${user} | [Calendar](${interviewer['link']})\n\n`;
   }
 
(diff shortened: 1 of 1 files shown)


src/commands/interviews/interviewers.ts
```diff
@@ -40,8 +40,7 @@ class InterviewersCommand extends BaseCommand {
   }
 
   private async getInterviewerDisplayInfo(interviewer: Interviewer) {
-    const userTag = (await this.client.users.fetch(interviewer['user_id'])).tag;
-    return '**' + userTag + '** | [Calendar](' + interviewer['link'] + ')\n\n';
+    return `${await this.client.users.fetch(interviewer['user_id'])} | [Calendar](${interviewer['link']})\n\n`;
```
gzcharleszhang, line 43: nit: can we take `await this.client.users.fetch(interviewer['user_id'])` out of the template string and store in a variable like before? I feel like it'd be more readable than having everything in 1 line.

gzcharleszhang (approved):

src/commands/interviews/interviewers.ts
Picowchew: Changed.

PR #117, opened 2021-11-15
Added ledger for Codey coin

### LINKED ISSUE #115: Create a ledger model and record codey coin balance changes

Picowchew:
# Related Issues
Resolves #115 

# Summary of Changes 
- Created the `user_coin_ledger` table with the following schema, noting that there is an index on `user_id`:
**id: number PK NOT NULL**
user_id: string NOT NULL
amount: number NOT NULL
new_balance: number NOT NULL
event: enum NOT NULL
reason: string
admin_id: string
created_at: datetime NOT NULL
- Removed the `updated_at` column from the `user_coin` table because this information can now be found in the `user_coin_ledger` table
- Made all of the timestamp columns in the tables not null
- Fixed coin-adjust's examples in its help message
- Minor grammar fix

# Steps to Reproduce
- Execute the SQL commands:
`ALTER TABLE user_coin DROP COLUMN updated_at;`
`CREATE TABLE suggestions2 (id INTEGER PRIMARY KEY NOT NULL, author_id VARCHAR(255) NOT NULL, author_username TEXT NOT NULL, created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP, suggestion TEXT NOT NULL, state VARCHAR(255) NOT NULL);`
`INSERT INTO suggestions2 SELECT * FROM suggestions;`
`DROP TABLE suggestions;`
`ALTER TABLE suggestions2 RENAME TO suggestions;`
`CREATE TABLE user_coin_bonus2 (user_id VARCHAR(255) NOT NULL, bonus_type INTEGER NOT NULL, last_granted TIMESTAMP NOT NULL, PRIMARY KEY (user_id, bonus_type));`
`INSERT INTO user_coin_bonus2 SELECT * FROM user_coin_bonus;`
`DROP TABLE user_coin_bonus;`
`ALTER TABLE user_coin_bonus2 RENAME TO user_coin_bonus;`
- Try various actions that cause a change to a user's Codey coin balance, such as `.coin-adjust`, and see `user_coin_ledger` update

### DIFF
--- a/src/commands/coin/coinAdjust.ts
+++ b/src/commands/coin/coinAdjust.ts
@@ -1,12 +1,7 @@
 import { Message, User } from 'discord.js';
 import { CommandoClient, CommandoMessage } from 'discord.js-commando';
 import { AdminCommand } from '../../utils/commands';
-import {
-  adjustCoinBalanceByUserId,
-  getCoinBalanceByUserId,
-  updateUserCoinLedger,
-  UserCoinEvent
-} from '../../components/coin';
+import { adjustCoinBalanceByUserId, getCoinBalanceByUserId, UserCoinEvent } from '../../components/coin';
 
 class CoinBalanceCommand extends AdminCommand {
   constructor(client: CommandoClient) {
@@ -44,8 +39,13 @@ class CoinBalanceCommand extends AdminCommand {
   async onRun(message: CommandoMessage, args: { user: User; amount: number; reason: string }): Promise<Message> {
     const { user, amount, reason } = args;
     // Update coin balance
-    await adjustCoinBalanceByUserId(user.id, amount);
-    await updateUserCoinLedger(user.id, amount, UserCoinEvent.CoinAdjust, reason ? reason : null, message.author.id);
+    await adjustCoinBalanceByUserId(
+      user.id,
+      amount,
+      UserCoinEvent.AdminCoinAdjust,
+      reason ? reason : null,
+      message.author.id
+    );
     // Get new balance
     const newBalance = await getCoinBalanceByUserId(user.id);
     return message.reply(`${user.username} now has ${newBalance} Codey coins 🪙.`);
--- a/src/commands/coin/coinAdjust.ts
+++ b/src/commands/coin/coinAdjust.ts
@@ -17,8 +17,8 @@ class CoinBalanceCommand extends AdminCommand {
       memberName: 'coin-adjust',
       description: "Adjusts a user's Codey coin balance by an amount.",
       examples: [
-        `${client.commandPrefix}coin-update @Codey 100`,
-        `${client.commandPrefix}coin-update @Codey -100 Codey broke.`
+        `${client.commandPrefix}coin-adjust @Codey 100`,
+        `${client.commandPrefix}coin-adjust @Codey -100 Codey broke.`
       ],
       args: [
         {
--- a/src/commands/coin/coinAdjust.ts
+++ b/src/commands/coin/coinAdjust.ts
@@ -1,7 +1,12 @@
 import { Message, User } from 'discord.js';
 import { CommandoClient, CommandoMessage } from 'discord.js-commando';
 import { AdminCommand } from '../../utils/commands';
-import { adjustCoinBalanceByUserId, getCoinBalanceByUserId } from '../../components/coin';
+import {
+  adjustCoinBalanceByUserId,
+  getCoinBalanceByUserId,
+  updateUserCoinLedger,
+  UserCoinEvent
+} from '../../components/coin';
 
 class CoinBalanceCommand extends AdminCommand {
   constructor(client: CommandoClient) {
@@ -11,7 +16,10 @@ class CoinBalanceCommand extends AdminCommand {
       group: 'coin',
       memberName: 'coin-adjust',
       description: "Adjusts a user's Codey coin balance by an amount.",
-      examples: [`${client.commandPrefix}coin-update @Codey 100`, `${client.commandPrefix}coin-update @Codey -100`],
+      examples: [
+        `${client.commandPrefix}coin-update @Codey 100`,
+        `${client.commandPrefix}coin-update @Codey -100 Codey broke.`
+      ],
       args: [
         {
           key: 'user',
@@ -22,15 +30,22 @@ class CoinBalanceCommand extends AdminCommand {
           key: 'amount',
           prompt: 'enter the amount to adjust.',
           type: 'integer'
+        },
+        {
+          key: 'reason',
+          prompt: '',
+          type: 'string',
+          default: ''
         }
       ]
     });
   }
 
-  async onRun(message: CommandoMessage, args: { user: User; amount: number }): Promise<Message> {
-    const { user, amount } = args;
+  async onRun(message: CommandoMessage, args: { user: User; amount: number; reason: string }): Promise<Message> {
+    const { user, amount, reason } = args;
     // Update coin balance
     await adjustCoinBalanceByUserId(user.id, amount);
+    await updateUserCoinLedger(user.id, amount, UserCoinEvent.CoinAdjust, reason ? reason : null, message.author.id);
     // Get new balance
     const newBalance = await getCoinBalanceByUserId(user.id);
     return message.reply(`${user.username} now has ${newBalance} Codey coins 🪙.`);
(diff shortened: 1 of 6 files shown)


src/components/coin.ts
```diff
@@ -3,13 +3,20 @@ import _ from 'lodash';
 
 import { openDB } from './db';
 
+export enum UserCoinEvent {
+  CoinAdjust,
+  CoinUpdate,
```
gzcharleszhang, line 8: can we prefix with Admin for these two events?

src/components/coin.ts
```diff
@@ -72,3 +96,25 @@ export const adjustCoinBalanceByUserId = async (userId: string, amount: number):
     amount
   );
 };
+
+/*
+  reason is only applicable for admin commands and is optional.
+  adminId is only applicable for admin commands and is mandatory.
+*/
+export const updateUserCoinLedger = async (
```
gzcharleszhang, line 104: maybe rename to `createCoinLedgerEntry`, i think update might be misleading here

gzcharleszhang: function docstring should also explain what the function does on a high level

src/commands/games/blackjack.ts
```diff
@@ -160,7 +165,7 @@ class BlackjackCommand extends BaseCommand {
         return `You won **${amountDiff}** Codey coin(s) ${getEmojiByName('codeyLove')}, keep your win streak going!`;
       }
       // player tied with dealer
-      return `Tied! You didn't win or lose any Codey coin(s), try again!`;
+      return `Tied! You didn't win nor lose any Codey coin(s), try again!`;
```
gzcharleszhang, line 168: doesn't sound natural to me but aight

src/commands/coin/coinUpdate.ts
```diff
@@ -22,17 +30,32 @@ class CoinBalanceCommand extends AdminCommand {
...
-    const { user, amount } = args;
+  async onRun(message: CommandoMessage, args: { user: User; amount: number; reason: string }): Promise<Message> {
+    const { user, amount, reason } = args;
+    // Get old balance
+    const oldBalance = await getCoinBalanceByUserId(user.id);
     // Update coin balance
     await updateCoinBalanceByUserId(user.id, amount);
     // Get new balance
     const newBalance = await getCoinBalanceByUserId(user.id);
+    await updateUserCoinLedger(
```
gzcharleszhang, line 52: adding entry to the ledger here is not a good idea, you should add entries to the ledger in the two update/adjust backend APIs, since this backend logic. you can just pass down an optional reason param to those functions if needed.

src/commands/coin/coinAdjust.ts
```diff
@@ -22,15 +30,22 @@ class CoinBalanceCommand extends AdminCommand {
           key: 'amount',
           prompt: 'enter the amount to adjust.',
           type: 'integer'
+        },
+        {
+          key: 'reason',
+          prompt: '',
```
gzcharleszhang, line 36: prompt?

src/commands/coin/coinUpdate.ts
```diff
@@ -22,17 +30,32 @@ class CoinBalanceCommand extends AdminCommand {
           key: 'amount',
           prompt: 'enter the new balance.',
           type: 'integer'
+        },
+        {
+          key: 'reason',
+          prompt: '',
```
gzcharleszhang, line 36: prompt?

src/commands/coin/coinAdjust.ts
```diff
@@ -22,15 +30,22 @@ class CoinBalanceCommand extends AdminCommand {
...
     });
   }
 
-  async onRun(message: CommandoMessage, args: { user: User; amount: number }): Promise<Message> {
-    const { user, amount } = args;
+  async onRun(message: CommandoMessage, args: { user: User; amount: number; reason: string }): Promise<Message> {
+    const { user, amount, reason } = args;
     // Update coin balance
     await adjustCoinBalanceByUserId(user.id, amount);
+    await updateUserCoinLedger(user.id, amount, UserCoinEvent.CoinAdjust, reason ? reason : null, message.author.id);
```
gzcharleszhang, line 48: same comment as before, not a good idea to update ledger here

src/commands/coin/coinAdjust.ts
Picowchew: This argument is optional as it has a default value. As such, even when this prompt is not an empty string, the prompt does not appear. Commando gives an error though if the prompt is removed, since it says that a prompt is required, so I just left it as an empty string.

src/commands/coin/coinUpdate.ts
Picowchew: Same as the comment above.

src/commands/coin/coinAdjust.ts
gzcharleszhang: makes sense!

src/components/coin.ts
```diff
@@ -3,16 +3,24 @@ import _ from 'lodash';
...
 }
 
-export type Bonus = { type: BonusType; amount: number; cooldown: number };
+export type Bonus = { type: BonusType; amount: number; cooldown: number; event: number };
 
 export const coinBonusMap = new Map<BonusType, Bonus>([
-  [BonusType.Daily, { type: BonusType.Daily, amount: 50, cooldown: 86400000 }], // one day in milliseconds
-  [BonusType.Activity, { type: BonusType.Activity, amount: 1, cooldown: 60000 }] // one minute in milliseconds
+  [BonusType.Daily, { type: BonusType.Daily, amount: 50, cooldown: 86400000, event: UserCoinEvent.BonusDaily }], // one day in milliseconds
+  [BonusType.Activity, { type: BonusType.Activity, amount: 1, cooldown: 60000, event: UserCoinEvent.BonusActivity }] // one minute in milliseconds
```
gzcharleszhang, line 23: 🧠

src/components/coin.ts
```diff
@@ -90,6 +129,31 @@ export const adjustCoinBalanceByUserId = async (userId: string, amount: number):
     amount,
     amount
   );
+  const newBalance = await getCoinBalanceByUserId(userId);
+  await createCoinLedgerEntry(userId, newBalance - oldBalance, event, reason, adminId);
```
gzcharleszhang, line 133: is it possible to use `amount` instead of calculating the diff using newBalance and oldBalance here?

Picowchew: I don't think so. As can be seen in the SQL command, the max is taken with 0, to take into account a negative adjustment with too large of a magnitude.

gzcharleszhang: If we need to get the old amount anyways, I think we can just calculate the new balance and re-use the update function to simplify logic. What do you think?

Picowchew: Won't calculating the new balance need the old balance? In that case, won't it be simpler to just call `createCoinLedgerEntry` instead of `updateCoinBalanceByUserId`?

gzcharleszhang: Hmm I see what you mean, I was trying to find a way to narrow the write down to one code path but I guess we need to get old balance either way.

We can keep the create ledger entry logic here then, but we can pre-compute the amount diff in memory? So we can avoid an extra DB call, and it makes the SQL simpler too. Also don't forget to add TODOs to wrap these in a transaction.

src/components/coin.ts
```diff
@@ -72,23 +136,31 @@ export const updateCoinBalanceByUserId = async (userId: string, newBalance: numb
...
-  The user's balance will be set to 0 if the adjustment brings it below 0.
+  Adds an entry to the Codey coin ledger due to a change in a user's coin balance.
+  reason is only applicable for admin commands and is optional.
+  adminId is only applicable for admin commands and is mandatory.
 */
-export const adjustCoinBalanceByUserId = async (userId: string, amount: number): Promise<void> => {
+export const createCoinLedgerEntry = async (
+  userId: string,
+  oldBalance: number,
+  updateAmount: number,
```
gzcharleszhang, line 150: nit: `updateAmount` confused me, I initially thought this is the amount diff or delta. Maybe rename to `newBalance`? Since we already have an `oldBalance` param here, which makes a lot of sense.

gzcharleszhang (approved):
1 nit, otherwise looks good!

PR #123, opened 2021-12-04
Coffee single match

### LINKED ISSUE #23: [EPIC] Coffee Chat Feature
**Tasks**
- [x] Write/translate algorithm that matches coffee chat members.
- [x] Handle the case where all matches have met each other at least once before.
- [x] DM matches their match result and write a short introduction.
- [x] Provides mods a summary of match results, preferably showing the number of times the pair has met before.
- [x] Improve documentation and readability of the matching algorithm.
- [x] Run simulations to get data and gain more confidence in the algorithm.
- [x] Implement a mod command to start the matching process and send the mod match results for approval.
- [ ] #83

**Is your feature request related to a problem? Please describe.**
The problem is finding a way to add more 1:1 connections in our community that is external to the executive committees. I want to be able to meet other people and have deeper conversations with them than speed-friending/afterhours events.

**Describe the solution you'd like**
Not sure if this is a command, but it would be nice to have something working like the Donut bot in Slack, where we can run it every [week/month], the opted-in people (through emojis or buttons probably) would be able to coffee chat with someone new. The bot would match them and maybe send each person the Discord id + tag so they can meet up.

I feel like this has been done by some Discord bot somewhere, so probably need a quick search first.

If not feasible, it's fine, I just thought this would be a cool thing to have on the Discord.

**Additional context**
https://www.donut.com/ for those that don't know what I'm talking about.

alexz429:
# Related Issues
resolves #23 

# Summary of Changes 
Added coffeematch and coffeetest functions for admins
Added corresponding database and .env dependencies

# Steps to Reproduce
Checkout branch
Add COFFEE_ROLE_ID which is the ID of the role to match coffee chatters with
run `.coffeematch` to send matches to everyone currently enrolled.
run `.coffeetest x` to simulate running match for x people.


src/commands/coffeechats/coffeeMatch.ts
```diff
@@ -0,0 +1,40 @@
...
+  async onRun(message: CommandoMessage): Promise<Message> {
+    //makes sure future matches are valid (made for the current group / still has matches left)
+    const { matches, single } = await getMatch(this.client);
+    for (const pair of matches) {
+      await this.alertMatch(pair[0], pair[1]);
+    }
+    await writeHistoricMatches(matches);
+    if (single) {
+      const singleUser = await this.client.users.fetch(single);
+      await message.reply(`${singleUser} is single and ready to mingle.`);
```
gzcharleszhang, line 27: what are we going to do with the single

src/components/coffeechat.ts
```diff
@@ -0,0 +1,273 @@
...
+  }
+  return { matches: matches, single: single };
+};
+
+/*
+ * Gets the list of users that are currently "enrolled" in coffee chat
+ * Returns a mapping of string -> int, where string is their ID, while int is an index assigned to the ID
+ * The index is used in place of the ID for match tallying
+ */
+const loadNotMatched = async (client: Client): Promise<Map<string, number>> => {
```
gzcharleszhang, line 63: nit: function name confused me, maybe `loadCoffeeChatUsers`?

src/components/coffeechat.ts
```diff
@@ -0,0 +1,273 @@
+import { Client } from 'discord.js-commando';
+import { Database } from 'sqlite';
+import { openDB } from './db';
+import _ from 'lodash';
+import { Person, stableMarriage } from 'stable-marriage';
+
+const COFFEE_ROLE_ID: string = process.env.COFFEE_ROLE_ID || '.';
+const TARGET_GUILD_ID: string = process.env.TARGET_GUILD_ID || '.';
```
gzcharleszhang, line 8: don't forget to include these in all the `.env` files. An alternative would be to create config files in the `config` directory, and specify these values for each environment. For example, in `config/production/vars.json`, add the variables for the production environment. Since role ID and guild ID are not sensitive info, I think it'd be ok for them to be part of the code.

src/components/coffeechat.ts
```diff
@@ -0,0 +1,273 @@
...
+  for (let i = 0; i < RANDOM_ITERATIONS; i++) {
+    //shuffle users, then separate into 2 sections
+    const notMatched = _.shuffle(Array.from(userList).map((name) => name[0]));
+    const A = notMatched.slice(0, Math.floor(notMatched.length / 2)).map((name) => new Person(name));
+    const B = notMatched.slice(Math.floor(notMatched.length / 2)).map((name) => new Person(name));
+
+    //attach preference list for each user by creating comparator which prioritizes user with less chats than the subject
+    A.forEach((value) =>
+      value.generatePreferences(
+        [...B].sort(
```
gzcharleszhang, line 162: This is getting hard to read, is it possible to factor out the following function? Maybe it's also reusable to generate preferences for `B`?

gzcharleszhang (commented):
Left some comments, generally looks good!

config/production/vars.json
```diff
@@ -0,0 +1,4 @@
+{
+  "TARGET_GUILD_ID": "701335585272627200",
```
gzcharleszhang, line 2: i think we need a config file for staging as well

src/components/coffeechat.ts
```diff
@@ -0,0 +1,265 @@
+import { Client } from 'discord.js-commando';
+import { Database } from 'sqlite';
+import { openDB } from './db';
+import vars from '../../config/production/vars.json';
```
gzcharleszhang, line 4: we shouldn't import from production by default, we should use `process.env.NODE_ENV` to determine which config file we pull

src/commands/coffeechats/coffeeMatch.ts
```diff
@@ -0,0 +1,50 @@
...
+    matches.forEach((pair) => {
+      if (!outputMap.get(pair[0])) outputMap.set(pair[0], []);
+      if (!outputMap.get(pair[1])) outputMap.set(pair[1], []);
+      outputMap.get(pair[0])!.push(pair[1]);
+      outputMap.get(pair[1])!.push(pair[0]);
+    });
+    //send out messages
+    outputMap.forEach(async (targets, user) => {
+      const discordUser = await this.client.users.fetch(user);
+      //we use raw discord id ping format to minimize fetch numbers on our end
```
gzcharleszhang, line 39: how does using discord ID ping format minimize fetches, is the alternative just using the `discordUser` object?

src/components/coffeechat.ts
```diff
@@ -0,0 +1,265 @@
...
+      ','
+    )};`,
+    _.flatten(newMatches)
+  );
+};
+
+/*
+ * Wipes a database
+ */
+// const wipeHistory = async (dbName: string): Promise<void> => {
```
gzcharleszhang, line 121: can we remove this if this is not needed anymore?

gzcharleszhang (commented):
A few comments/questions, otherwise looks good

config/production/vars.json
gzcharleszhang: possibly for development too

src/commands/coffeechats/coffeeMatch.ts
alexz429: basically just saves us an await call, since clients.users.fetch is async.

Alternatively, we would have to build a dict of all the id's so that we don't have multiple instances of the same discord User in the map.

gzcharleszhang: oh I see what this is doing, I think this is fine as long as the @mention won't just show up as an ID string as that's bad user experience

src/bot.ts
```diff
@@ -12,8 +12,11 @@ import logger, { logError } from './components/logger';
 import { messageListener } from './components/messageListener';
 import { initEmojis } from './components/emojis';
 import { createSuggestionCron, createBonusInterviewerListCron } from './components/cron';
+import { readFileSync } from 'fs';
 
-const NOTIF_CHANNEL_ID: string = process.env.NOTIF_CHANNEL_ID || '.';
+const ENVIRONMENT: string = process.env.ENVIRONMENT || '.';
```
gzcharleszhang, line 17: i think the var you're looking for here is `NODE_ENV`?

src/bot.ts
```diff
@@ -12,8 +12,11 @@ import logger, { logError } from './components/logger';
...
+const vars = JSON.parse(readFileSync(`./config/${ENVIRONMENT}/vars.json`, 'utf-8'));
```
gzcharleszhang, line 18: be careful there because the NODE_ENV for dev is specified as `dev`

src/commands/coffeechats/coffeeMatch.ts
alexz429: For some reason this seems to show ID's occasionally, and I can't figure out whether its a discord caching issue or something else entirely. I think a safer bet is to just refactor this to use the fetch properly.

gzcharleszhang (approved):
Logic looks good, we can fine-tune UX stuff in subsequent PRs.

PR #124, opened 2021-12-14
Daily Codey coin bonus to those on interviewer list

### LINKED ISSUE #121: Codey coin bonus for being on the interviewer list
Interviewers on the interviewer list would passively receive a daily Codey coin bonus.

Picowchew:
# Related Issues
Resolves #121 

# Summary of Changes 
A daily Codey coin bonus is given to those on the interviewer list

# Steps to Reproduce
At 1 am UTC, see the bonus being given

### DIFF
--- a/src/components/cron.ts
+++ b/src/components/cron.ts
@@ -39,6 +39,9 @@ export const createBonusInterviewerListCron = (): CronJob =>
     const activeInterviewers = await getInterviewers(null);
     activeInterviewers.forEach(async (interviewer) => {
       const bonus = coinBonusMap.get(BonusType.InterviewerList);
-      await adjustCoinBalanceByUserId(interviewer.user_id, bonus!.amount, bonus!.event);
+      if (!bonus) {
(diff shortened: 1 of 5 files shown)


src/components/cron.ts
```diff
@@ -30,3 +32,13 @@ export const createSuggestionCron = (client: CommandoClient): CronJob =>
       }
     }
   });
+
+// Gives Codey coin bonus to those on the interviewer list at 1:00 am UTC
+export const createBonusInterviewerListCron = (): CronJob =>
+  new CronJob('0 0 1 * * *', async function () {
+    const activeInterviewers = await getInterviewers(null);
+    activeInterviewers.forEach(async (interviewer) => {
+      const bonus = coinBonusMap.get(BonusType.InterviewerList);
```
gzcharleszhang, line 41: what if `coinBonusMap` doesn't contain the desired bonus type? can we do a null check so we don't need non null assertions? I feel like non null assertions should be used sparingly.

gzcharleszhang (approved):
1 minor comment, otherwise LGTM

PR #125, opened 2022-01-02
Fixed interviewer list bonus being applied as activity bonus

Picowchew:
# Related Issues
N/A

# Summary of Changes 
Fixed the interviewer list bonus being attempted to be applied as an activity bonus, in which an error is thrown because the interviewer list bonus does not have a cooldown

# Steps to Reproduce
Send some messages and see that an error is not thrown anymore

### DIFF
--- a/src/components/coin.ts
+++ b/src/components/coin.ts
@@ -7,14 +7,14 @@ export enum UserCoinEvent {
   AdminCoinAdjust,
   AdminCoinUpdate,
   BonusDaily,
-  BonusMinute,
+  BonusActivity,
   BonusInterviewerList,
   Blackjack
 }
 
 export enum BonusType {
   Daily = 0,
-  Minute,
+  Activity,
   InterviewerList
 }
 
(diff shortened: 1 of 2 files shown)


src/components/coin.ts
```diff
@@ -7,29 +7,58 @@ export enum UserCoinEvent {
   AdminCoinAdjust,
   AdminCoinUpdate,
   BonusDaily,
-  BonusActivity,
+  BonusMinute,
```
gzcharleszhang, line 10: We didn't name this minute because we might change it to be more than a minute

src/components/coin.ts
```diff
@@ -7,29 +7,58 @@ export enum UserCoinEvent {
...
   InterviewerList
 }
 
-export type Bonus = { type: BonusType; amount: number; cooldown: number | null; event: number };
+export type Bonus = {
+  type: BonusType;
+  event: number;
+  amount: number;
+  cooldown: number | null;
+  isActivity: boolean;
```
gzcharleszhang, line 26: maybe isMessageBonus? And add a comment for this field too

gzcharleszhang (approved):
Minor comments, otherwise LGTM

Issue #95, opened 2021-10-21
Rolling a dice


gzcharleszhang:
A `.dice` command to roll a 6-sided dice and reply to the user the side the dice landed on. Each side should have an equal probability of getting chosen.

Bonus: accept an integer parameter `n` to roll an `n`-sided dice.

PR #136, opened 2022-01-29
Roll dice command

EdwinYi-JanYang:
# Related Issues
Resolves Issue #95 

# Summary of Changes 
Adds command .rd n which rolls an n-sided dice and pings the user with the returned value.

# Steps to Reproduce

### DIFF
--- a/src/commands/fun/rollDice.ts
+++ b/src/commands/fun/rollDice.ts
@@ -22,7 +22,7 @@ class RollDiceCommand extends BaseCommand {
   }
 
   getRandomInt(max: number): number {
-    const roll = Math.floor(Math.random() * max);
+    const roll = Math.floor(Math.random() * (max + 1));
     if (roll == 0) {
       return 1;
     }
@@ -34,7 +34,7 @@ class RollDiceCommand extends BaseCommand {
     if (sides <= 0) {
       return message.reply(`I cannot compute ` + sides + ' sides!');
     }
-    else if (sides > 1000000) {
+    if (sides > 1000000) {
       return message.reply(`that's too many sides!`);
     }
     const diceFace = this.getRandomInt(sides);
--- a/src/commands/fun/rollDice.ts
+++ b/src/commands/fun/rollDice.ts
@@ -22,13 +22,17 @@ class RollDiceCommand extends BaseCommand {
   }
 
   getRandomInt(max: number): number {
-    return Math.floor(Math.random() * max);
+    const roll = Math.floor(Math.random() * max);
+    if (roll == 0) {
+      return 1;
+    }
+    return roll;
   }
 
   async onRun(message: CommandoMessage, args: {sides: number;}): Promise<Message> {
     const {sides} = args;
     if (sides <= 0) {
-      return message.reply(`cannot compute ` + sides + ' sides!');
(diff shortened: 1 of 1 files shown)


src/commands/fun/rollDice.ts
```diff
@@ -0,0 +1,45 @@
...
+
+class RollDiceCommand extends BaseCommand {
+  constructor(client: CommandoClient) {
+    super(client, {
+      name: 'roll-dice',
+      aliases: ['rd', 'roll', 'rolldice', 'dice-roll', 'diceroll', 'dice'],
+      group: 'fun',
+      args: [
+        {
+          key: 'sides',
```
gzcharleszhang, line 13: Can default to 6 sides

src/commands/fun/rollDice.ts
```diff
@@ -0,0 +1,45 @@
...
+          type: 'integer'
+        }
+      ],
+      memberName: 'dice-coin',
+      description: 'Roll a dice!',
+      examples: [`${client.commandPrefix}dice-roll`]
+    });
+  }
+
+  getRandomInt(max: number): number {
```
gzcharleszhang, line 24: I think this gives you [1, max) instead of [1, max]

src/commands/fun/rollDice.ts
```diff
@@ -0,0 +1,45 @@
...
+    }
+    return roll;
+  }
+
+  async onRun(message: CommandoMessage, args: {sides: number;}): Promise<Message> {
+    const {sides} = args;
+    if (sides <= 0) {
+      return message.reply(`I cannot compute ` + sides + ' sides!');
+    }
+    else if (sides > 1000000) {
```
gzcharleszhang, line 37: Can change else if to an if since the previous of statement exits the function

gzcharleszhang (changes requested):

src/commands/fun/rollDice.ts
EdwinYi-JanYang: how would I default to 6 sides?

src/commands/fun/rollDice.ts
EdwinYi-JanYang: yes this is a problem

src/commands/fun/rollDice.ts
EdwinYi-JanYang: understood

src/commands/fun/rollDice.ts
EdwinYi-JanYang: it should be fixed now

src/commands/fun/rollDice.ts
gzcharleszhang: Try adding default: 6

EdwinYi-JanYang: alright

src/commands/fun/rollDice.ts
gzcharleszhang: This is still not quite right, the probability of rolling a 1 is higher than all other options

EdwinYi-JanYang: This is very bad

gzcharleszhang (approved):

Issue #75, opened 2021-09-26
Use GCP cloud logging driver for each codey container


gzcharleszhang:
Currently, we have two types of logs:
1. Default docker container logs (stdout and stderr), which are saved inside `/var/lib/docker/containers/.../*-log.json`.
2. Command level logs using winston, which are saved in the `logs` directory for each codey deployment.

This poses a few problems:
- Hard to query logs, need to ssh into machine.
- Logs are not retained outside of the GCP instance.
- Logs are not structured.
- Logs are stored in two locations.

We should look into using the GCP log driver for our Codey containers, so we can use GCP cloud logging query viewer to see the logs inside GCP console.

Alluq:
QQ, enabling cloud logging requires us to set up billing and everything, which might cost some money depending on how many logs we have, the first 50 gbs of logs are free but afterwards its 0.5 USD per GB. 
Are we going to reach that threshold? If so do we need to worry about funding for the codey bot?

gzcharleszhang:
I don't think we will reach that threshold. We can enable cloud logging and monitor for usage for 1-2 weeks. Even if we go a little bit extra, we do have budget for the bot.

Alluq:
you can now query for the logs of staging and production by adding this to the query text box in the log menu:
![image](https://user-images.githubusercontent.com/39013427/153137949-113d25ac-3c30-4f8a-a08d-95178178e557.png)
`jsonPayload.container.name: <name of codey instance here>`
for example with staging you can find all staging logs using:
`jsonPayload.container.name:/codey-staging`

PR #151, opened 2022-02-02
Using GCP log driver in staging

Alluq:
# Related Issues
Hopefully resolves the issue #75 in staging

# Summary of Changes 
Very small change to the docker-compose.yml basically just added the logging option according to https://cloud.google.com/community/tutorials/docker-gcplogs-driver

If this works for staging then 

# Steps to Reproduce
Go to docker-compose.yml in staging and add 
```
logging: 
  driver: gcplogs
```
Into the codey-bot service


docker-compose.yml
```diff
@@ -10,3 +10,5 @@ services:
       - ./src:/usr/app/src
       - ./logs:/usr/app/logs
       - ./db:/usr/app/db
+    logging:
+      driver: gcplogs
```
gzcharleszhang, line 14: I think this is for running docker containers locally, so we prob don't want to use the gcp driver here

gzcharleszhang (changes requested):

docker-compose.yml
Alluq: okay yeah I'll remove it from local docker-compose

gzcharleszhang (approved):

PR #152, opened 2022-02-03
Added spam and troll detection

### LINKED ISSUE #138: Spam and troll detection
Auto mute people for 1 hour and 1 minute who:
- consecutively made the same 3 messages in the 10 most recent messages
- try to @ everyone with a link OR "nitro" instance outside of PC

Picowchew:
# Related Issues
Resolves #138 

# Summary of Changes 
If someone does @ everyone or @ here and includes a link with "http" or "nitro" outside of the PC category or if someone sends a message in the honeypot channel, then they get kicked and their relevant messages get deleted. The kick gets logged.

# Steps to Reproduce
Try the above


src/components/messageListener.ts
```diff
@@ -2,8 +2,36 @@ import { Message } from 'discord.js';
...
+    (pingWords.some((word) => message.content.includes(word)) &&
+      punishableWords.some((word) => message.content.toLowerCase().includes(word)) &&
+      message.channel instanceof TextChannel &&
+      message.channel.parentID !== PC_CATEGORY_ID) ||
+    message.channel.id === HONEYPOT_CHANNEL_ID
+  ) {
+    // Soft bans member and deletes messages from past 1 day
+    try {
+      if (await message.member?.ban({ days: 1, reason: 'Spammer/troll/got hacked' })) {
+        await message.guild?.members.unban(message.author.id);
```
gzcharleszhang, line 27: Just curious, is there a reason why we need to wrap this in an if statement, if `ban` fails, wouldn't it throw an error and skip `unban`? I.e. is there a case where `ban` returns `false` instead of throwing an error.

gzcharleszhang (changes requested):
really cool stuff, left some comments about code style

src/components/messageListener.ts
Picowchew: In the if condition, `message.member` might be null, in which undefined would returned, which would evaluate to false. I guess in this case though, `message.member` would never be null based on the if statement above it, but I guess this *might* change in the future?

marko-polo-cheno: Can you test out a code snippet that would trigger spam detection? like ```@everyone give nitro``` 
Also, plz ban for 1 days and 1 hour in case someone auto spams exactly and every 24 hours.
Do we have protection against an exec or mod or advisor or someone of somewhat higher CSC status accidently getting banned and all msgs deleted?

marko-polo-cheno (changes requested):
Also constants -> config if not done already :)
Rest looks good to me - ty Charles u beat me to it.

src/components/messageListener.ts
Picowchew: Yup, the inline code block also triggers spam detection.
I think a soft ban in general is a decent punishment, since there's the added difficulty of having to rejoin the server before potentially spamming again. If a temporary ban would be implemented though, then would this involve storing the time of unban in the database and a cron job to check if it is time to do any unbans?
The Executive, Discord Mod, and Advisor roles are above the Codey role, so Codey can't possibly ban anyone with these roles.

marko-polo-cheno (approved):
is there protection for general PC members?
would be good cause not everyone is exec, mod, or advisor

config/config.ts
```diff
@@ -0,0 +1,4 @@
+import { readFileSync } from 'fs';
```
gzcharleszhang, line 1: nit: this file can be be under `/src/config.ts`. IMO the `/config` directory should only contain raw config data, how the app reads/interprets that config data should be under the app's own directory aka `/src`

src/components/messageListener.ts
```diff
@@ -2,8 +2,53 @@ import { Message } from 'discord.js';
...
+
+/*
+ * Punish spammers/trolls/people who got hacked
+ */
+const punishSpammersAndTrolls = async (message: Message): Promise<boolean> => {
+  if (detectSpammersAndTrollsNotByHoneypot(message) || message.channel.id === HONEYPOT_CHANNEL_ID) {
+    // Temp ban member for 1 day and 1 hour, and delete messages from past 1 day
+    try {
+      if (await message.member?.ban({ days: 1, reason: 'Spammer/troll/got hacked' })) {
+        setTimeout(async () => await message.guild?.members.unban(message.author.id), 25 * 60 * 60 * 1000); // 1 day and 1 hour in milliseconds
```
gzcharleszhang, line 37: This implementation may not be resilient. Let's say if a user gets banned, and then the bot crashes or a new version gets deployed within 24 hours, then the user will not get unbanned.

src/components/messageListener.ts
```diff
@@ -2,8 +2,53 @@ import { Message } from 'discord.js';
...
+  return (
+    pingWords.some((word) => message.content.includes(word)) &&
+    punishableWords.some((word) => message.content.toLowerCase().includes(word)) &&
+    message.channel instanceof TextChannel &&
+    message.channel.parentID !== PC_CATEGORY_ID
+  );
+};
+
+/*
+ * Punish spammers/trolls/people who got hacked
```
gzcharleszhang, line 30: Mention what the return values mean, i.e. true means ..., and false otherwise

src/components/messageListener.ts
Picowchew: An alternative solution that I can think of is that the time of unban would be stored in the database and there would be a cron job to check if it is time to do any unbans? Also, is the current implementation okay or is a more resilient implementation wanted?

Picowchew:
> is there protection for general PC members? would be good cause not everyone is exec, mod, or advisor

Yup, similarly, the PC role is above the Codey role, so Codey can't ban people with the PC role. I guess in the case that the roles get moved around, then the PC category check would help. For the official documentation, according to https://discord.com/developers/docs/topics/permissions, "A bot can only kick, ban, and edit nicknames for users whose highest role is lower than the bot's highest role."

src/components/messageListener.ts
marko-polo-cheno: A db isn't needed as long as we print out exactly who was banned in # moderation and when they were banned, prior to actually kicking them.

We can manually do unbans upon request, which is probably very rare anyways.
What if we KICKED instead of banning for some amt of time? Right now, I think msgs disappear, which might not always be good. Suppose I stopped being a mod, and became a norm member, and got hacked, and made spam, hence banned - then would my 4000+ msgs all disappear?

src/components/messageListener.ts
```diff
@@ -2,8 +2,53 @@ import { Message } from 'discord.js';
...
+ * Return true if someone of this kind is detected, false otherwise
+ */
+const punishSpammersAndTrolls = async (message: Message): Promise<boolean> => {
+  if (detectSpammersAndTrollsNotByHoneypot(message) || message.channel.id === HONEYPOT_CHANNEL_ID) {
+    // Delete the message, and kick the member if they are still in the server
+    try {
+      await message.delete();
+      await message.member?.kick('Spammer/troll/got hacked');
```
gzcharleszhang, line 38: do we want to send the mod channel some kind of message? so we know someone was kicked

gzcharleszhang (approved):
1 minor comment otherwise LGTM

src/components/messageListener.ts
marko-polo-cheno: yes
``A db isn't needed as long as we print out exactly who was kicked in # moderation and the msg they were kicked for, and when they were banned, prior to actually kicking them.``

marko-polo-cheno: Also :codey-poggers: to @Picowchew for getting this done so fast!

Picowchew: Yup done, except that I did the logging after the kick, since I don't think we want there to be logging if the kick doesn't go through (e.g. Codey cannot kick a mod).

README.md
```diff
@@ -9,7 +9,6 @@ Set these accordingly within the correct config folder. If you are testing local
 - `TARGET_GUILD_ID`: the guild (server) in which coffee chats are being held.
 - `COFFEE_ROLE_ID`: the ID of the role the bot will use to decide who is enrolled into coffeechats.
 - `NOTIF_CHANNEL_ID`: the ID of the channel the bot will send system notifications to.
-- `MOD_CHANNEL_ID` : the ID of the discord mod channel.
```
marko-polo-cheno, line 12: plz make sure honey id is in readme if not already there

marko-polo-cheno (approved):
great stuff! :D

marko-polo-cheno:
wait
one more thing, maybe it's good to log triggered kicks, even if they don't have an effect on mods

marko-polo-cheno (approved):
awesome :) i guess u can merge and tell ppl to pull latest since there are some changes that effect a lot of files

Issue #139, opened 2022-01-30
Create db user_profile


marko-polo-cheno:
Create a table called user_profile using discord ids as the primary key.
Add relevant columns such as about me, MM-YY of birth, preferred name, preferred pronouns, etc.

PR #155, opened 2022-02-05
Create db user profile

smilesthesad:
# Related Issues
resolves <https://github.com/uwcsc/codeybot/issues/139>

# Summary of Changes 
Added user profile database

# Steps to Reproduce
Create the database, try inserting things in


src/components/profile.ts
```diff
@@ -0,0 +1,20 @@
+import { Database } from 'sqlite';
+export const initUserProfileTable = async (db: Database): Promise<void> => {
```
gzcharleszhang, line 2: I think @Picowchew recently moved all the init table stuff to `db.ts`, so we should follow that convention too

src/components/db.ts
```diff
@@ -1,5 +1,12 @@
 import sqlite3 from 'sqlite3';
 import { open, Database } from 'sqlite';
+
+import { initSuggestionsTable } from './suggestions';
```
marko-polo-cheno, line 4: ya i may have done smth here to fix a conflict, but plz pull latest bc the format of db.ts changed to avoid cyclic imports

marko-polo-cheno (changes requested):

marko-polo-cheno (approved):
LGTM :)
The only thing I would suggest changing is ``user_table`` to ``user_profile_table`` bc 'profile' is more descriptive compared to 'user'.

Keep in mind that the fields in ``user_table`` are bound to change in the future.
Feel free to merge as well as self-assign another ticket. Do lmk if u ever need more details tho.

PR #167, opened 2022-02-10
Modified message listener so bot ignores its own messages

Picowchew:
# Related Issues
N/A

# Summary of Changes 
- Modified the message listener so Codey always ignores its own messages; this would allow Codey to send a message in the honeypot channel without getting its own message deleted (and also help prevent infinite loops in general)
- Also allow a user to receive a coin bonus if their message is sent in the announcements channel

# Steps to Reproduce
One way to test this is to make sure Codey's coin balance does not go up after sending messages


marko-polo-cheno (approved):
o smart moves

PR #174, opened 2022-02-24
Migrated Codey coin to Sapphire

### LINKED ISSUE #159: migrate codeycoin to sapphire

Picowchew:
# Related Issues
Resolves #159 

# Summary of Changes 
- Migrated Codey coin to Sapphire
- Made coin check command available to everyone

# Steps to Reproduce
Try the coin commands


marko-polo-cheno (approved):
looks pretty good!
do we need to delete oldCode? or can we keep it until we know all this code works the way we want it to?
Is there a way to test sapphire stuff rn (to get screenshots as proof of it working) to include in PRs?

Picowchew:
I think removing the relevant files from the oldCode folder helps with knowing what has already been migrated? Yup! The branch works; just make sure to install the dependencies and update the .env file as per README.md. Below is a sample screenshot.
image

src/commands/coin/coinCheck.ts
```diff
@@ -0,0 +1,24 @@
...
+import { Args, Command, CommandOptions } from '@sapphire/framework';
+import { ApplyOptions } from '@sapphire/decorators';
+import { getCoinBalanceByUserId } from '../../components/coin';
+import { BOT_PREFIX } from '../../bot';
+
+@ApplyOptions<CommandOptions>({
+  aliases: ['cc', 'coin-balance-check', 'coin-check'],
+  description: "Checks a user's Codey coin balance.",
+  detailedDescription: `**Examples:**\n\`${BOT_PREFIX}coin-check @Codey\`\n\`${BOT_PREFIX}cc @Codey\``,
+  requiredUserPermissions: ['ADMINISTRATOR']
```
smilesthesad, line 11: Why does this require admin? We're displaying user coins on their profile as well so I think this command should be available to everyone?

Picowchew: It was discussed during a meeting in the previous term that a user may not want their coin balance to be checked by other people, so this command was made admin-only. If someone wants to check someone else's balance, then they would have to ask the person to do `.bal` in a public channel.
I did not realize that coin balance would be displayed on a user profile. Are we sure that this is a change that is wanted? @marko-polo-cheno

marko-polo-cheno: @smilesthesad @Picowchew good point
imo we should open it up to public, but let's talk abt it on Wed.
if public, then we can have a leaderboard command :o

alexz429 (approved):
looks all good, but lets discuss the admin perm on wed

PR #176, opened 2022-03-02
Migrate fun to sapphire

Alluq:
# Related Issues
Resolves #163 

# Summary of Changes 
I changed the rollDice and flipCoin to work with sapphire's classes and turned their constructors into decorators for brevity.

# Steps to Reproduce
Test and try out the flip coin commands and the roll dice commands

PR #179, opened 2022-03-15
Reduced file dependency and enforced 0 arguments

Picowchew:
# Related Issues
N/A

# Summary of Changes 
- Client is now obtained from container; this reduces file dependency since previously, the client had to be imported from bot.ts, and this also allows for the client to not be needed as a parameter anymore
- Augmented Container to have the botPrefix property, since container.botPrefix is shorter than container.client.options.defaultPrefix; this also reduces the file dependency from importing the bot prefix from bot.ts
- Enforced 0 arguments for commands that don't take any arguments
- Updated setFooter of MessageEmbed to discord.js v13
- Organized imports
- Fixed some typos

# Steps to Reproduce
Check that there is enforcement for 0 arguments, and other than that, there should be no apparent change

### DIFF
--- a/src/utils/arguments.ts
+++ b/src/utils/arguments.ts
@@ -1,0 +1,3 @@
+export const enforceNoArgumentsMessage = (commandName: string): string => {
+  return `\`${commandName.toLowerCase()}\` doesn't take any arguments.`;
+};
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -1,11 +1,10 @@
 import dotenv from 'dotenv';
 dotenv.config();
 
+import { container, LogLevel, SapphireClient, SapphirePrefix } from '@sapphire/framework';
 import '@sapphire/plugin-logger/register';
 import * as colorette from 'colorette';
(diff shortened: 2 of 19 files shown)


marko-polo-cheno (approved):

src/utils/arguments.ts
```diff
@@ -0,0 +1,3 @@
+export const enforceNoArgumentsMessage = (commandName: string): string => {
+  return `\`${commandName.toLowerCase()}\` doesn't take any arguments.`;
```
marko-polo-cheno, line 2: If I do `.balance 234` or whatever, it'll say dont give args. Why not run anyways?

Picowchew: Oh this was discussed in https://github.com/uwcsc/codeybot/pull/176#discussion_r823344516, in which the conclusion seems to be to enforce 0 arguments?

Picowchew: I have now removed 0 argument enforcement.

alexz429 (approved):

PR #196, opened 2022-04-11
Use subcommands

Picowchew:
# Related Issues
N/A

# Summary of Changes 
- Refactored coffee chat and coin to use subcommands
- Added aliases, including for the subcommands themselves (also added `.coin information`, since it's odd how `.coin info` works, but not `.coin information`)
- Made some fields private
- Updated some folder and file naming across commands and components for consistency
- Fixed there being extra whitespace (including newlines) in examples
- Capitalized Discord and ID, and added some periods, in suggestions
- Removed oldCode (suggestions, logger, and command utils) and removed some redundant code
- Organized imports
- Reformatted the code to adhere to Prettier 

# Steps to Reproduce
For the most part, the only apparent change would be more aliases


src/commands/coin/coin.ts
```diff
@@ -0,0 +1,142 @@
...
+    { input: 'bal', output: 'balance' },
+    'check',
+    { input: 'c', output: 'check' },
+    'information',
+    { input: 'info', output: 'information' },
+    { input: 'i', output: 'information' },
+    'update',
+    { input: 'u', output: 'update' },
+    { input: 'set', output: 'update' },
+    { input: 's', output: 'update' }
```
alexz429, line 40: oh wow u figured out aliasing good work

src/commands/games/blackjack.ts
```diff
@@ -27,15 +27,15 @@ const validateBetAmount = (amount: number): string => {
 @ApplyOptions<CommandOptions>({
   aliases: ['blj', 'bj'],
   description: 'Start a Blackjack game to win some Codey coins!',
-  detailedDescription: `**Examples:**\n
-  \`${container.botPrefix}blackjack 100\`\n
```
alexz429, line 31: what does removing the newlines actually do?

alexz429 (approved):

src/commands/games/blackjack.ts
Picowchew: So I actually tried printing out the strings this time and I realized that the fact that a line is written on a different line already means that there's a newline. With the `\n`, that caused double space instead of single space. The extra whitespace at the beginning of a line made some unwanted indents.

PR #218, opened 2022-07-02
lock sapphire framework version and remove deprecated options

alexz429:
Because slash commands is a newer feature for sapphire, it is heavily under development.

Until the dust settles down, we will be keeping sapphire framework version to `3.0.0-next.8198e94.0` unless there's some massive bug. 

Fixed deprecated options in slash command setup.


mcpenguin (approved):
LGTM!

PR #221, opened 2022-07-05
Slash commands - added registerApplicationCommands

mcpenguin:
As above.


src/commands/fun/flipCoin.ts
```diff
@@ -1,8 +1,13 @@
-import { ApplyOptions } from '@sapphire/decorators';
-import { Command, CommandOptions, container } from '@sapphire/framework';
+// import { ApplyOptions } from '@sapphire/decorators';
```
alexz429, line 1: might as well delete these

alexz429 (approved):
small nits and run the linter, otherwise good

PR #226, opened 2022-07-07
Removed honeypot

Picowchew:
# Related Issues

N/A

# Summary of Changes

Removed the honeypot functionality because the other method works well enough

# Steps to Reproduce

A person who types a message in the honeypot channel would not be kicked anymore


marko-polo-cheno (approved):
can u just add a comment on what exactly needs to be re-added if we did want honeypot to exist again?
ie. what to do with honeypot channel id

Issue #185, opened 2022-03-31
[FEATURE] Attach somebody's discord @ to ban messages for appeal channel


alexz429:
When somebody gets banned, they should need at least some form of way to communicate for appeals outside of exec email. No real functionality to log with codey has been done yet, but if they have one of the existing mods @ they can DM for appeals.

Would be nice to add this to config so its an easy change across years.

smilesthesad:
By "somebody" should we have an array of available mods in the config? Or just have one individual

alexz429:
Most likely one is sufficient, should just be an extra config var.

PR #235, opened 2022-07-12
added Moderator tag to ban message in messageCreate.ts

WWW0030:
# Related Issues

# Summary of Changes

# Steps to Reproduce

### DIFF
--- a/src/listeners/messageCreate.ts
+++ b/src/listeners/messageCreate.ts
@@ -7,6 +7,7 @@ import { sendKickEmbed } from '../utils/embeds';
 
 const HONEYPOT_CHANNEL_ID: string = vars.HONEYPOT_CHANNEL_ID;
 const ANNOUNCEMENTS_CHANNEL_ID: string = vars.ANNOUNCEMENTS_CHANNEL_ID;
+const APPEAL_MODERATOR_ID: string = "ǝʞıɹʇSǝldıɹ⊥#5402";
 
 /*
  * Detect spammers/trolls/people who got hacked, not by the honeypot method, by
@@ -38,7 +39,7 @@ const punishSpammersAndTrolls = async (message: Message): Promise<boolean> => {
     await message.delete();
     if (message.member) {
       const user = message.member.user;
-      const reason = 'Spammer/troll/got hacked';
+      const reason = 'Spammer/troll/got hacked'+'for appeal, please contact: '+APPEAL_MODERATOR_ID;
       let isSuccessful = true;
       try {
         await message.member.kick(reason);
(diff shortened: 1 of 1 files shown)


Picowchew (changes requested):
Good start! I have a few comments.

It would be nice if you could fill out the PR description. There are comments in the PR template to describe what is meant to be written in each section. This makes it much easier for the reviewer to go through your PR. For instance, I was actually confused at first as to the purpose of this PR; was it supposed to DM the banned person with the moderator for them to contact or was it supposed to log which moderator (or Codey) did the banning. I later received clarification that you are working on this ticket https://github.com/uwcsc/codeybot/issues/185, so in the "Related Issues" section, you can do something like `Resolves #185` to automatically close the issue when the PR is merged (https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).

I believe that what the code currently does is that in the log message (which is sent in the server) that is only accessible by mods and executives, the username and discriminator of a mod is added to the reason. What should be done instead is that the banned person should receive a DM from Codey with a ping to a mod. In this way, the banned person would be able to contact a mod after leaving the server due to the ban. The reason why a ping is used is that the banned person can easily click on the ping and then message the mod. If they do not have any mutual servers though, then from the ping, the username and discriminator can be obtained, and then from this information, the banned person is able to send the mod a friend request. After accepting the friend request, the banned person is able to message the mod. The coffee chat code DMs people with a ping, so feel free to look there for some guidance on how this could be done. In addition, as described in the issue, you can add the ID of the mod as another config variable as opposed to making the username and discriminator a constant within the file. This allows for easier changing of the mod.

Feel free to reply here if you have any questions!

src/listeners/messageCreate.ts
```diff
@@ -38,7 +39,7 @@ const punishSpammersAndTrolls = async (message: Message): Promise<boolean> => {
     await message.delete();
     if (message.member) {
       const user = message.member.user;
-      const reason = 'Spammer/troll/got hacked';
+      const reason = 'Spammer/troll/got hacked'+'for appeal, please contact: '+APPEAL_MODERATOR_ID;
```
alexz429, line 42: make sure this looks right when you test it, there doesn't seem to be a space between hacked and for

src/listeners/messageCreate.ts
```diff
@@ -7,6 +7,7 @@ import { sendKickEmbed } from '../utils/embeds';
 
 const HONEYPOT_CHANNEL_ID: string = vars.HONEYPOT_CHANNEL_ID;
 const ANNOUNCEMENTS_CHANNEL_ID: string = vars.ANNOUNCEMENTS_CHANNEL_ID;
```
alexz429, line 9: check these linting errors, typically if you installed a linter into your IDE these should be autofixed.

alexz429 (changes requested):

mcpenguin:
Any updates on this?

mcpenguin:
Closing this PR due to inactivity.

PR #236, opened 2022-07-14
Only edit message for ping

Picowchew:
# Related Issues

N/A

# Summary of Changes

Originally, non-slash commands used to send both the `messageWhenExecutingCommand` and the final response. Now, only `ping` would edit the message from `messageWhenExecutingCommand` to the final response and the other commands would only send the final response.

# Steps to Reproduce

Try out the commands
image

### DIFF
--- a/src/codeyCommand.ts
+++ b/src/codeyCommand.ts
@@ -47,13 +47,11 @@ export class CodeyCommand extends SapphireCommand {
   }
 
   // Regular command
-  public async messageRun(message: Message): Promise<Message<boolean>> {
+  public async messageRun(message: Message): Promise<Message<boolean> | undefined> {
     const { client } = container;
-    const initialMessageFromBot: SapphireMessageRequest = await message.channel.send({
-      content: this.messageWhenExecutingCommand
-    });
     try {
-      const successResponse = await this.executeCommand(client, message, initialMessageFromBot);
+      const successResponse = await this.executeCommand(client, message);
+      if (!successResponse) return;
       switch (this.codeyCommandResponseType) {
         case CodeyCommandResponseType.EMBED:
           return await message.channel.send({ embeds: [<MessageEmbed>successResponse] });
@@ -67,7 +65,9 @@ export class CodeyCommand extends SapphireCommand {
   }
 
   // Slash command
-  public async chatInputRun(interaction: SapphireCommand.ChatInputInteraction): Promise<APIMessage | Message<boolean>> {
+  public async chatInputRun(
+    interaction: SapphireCommand.ChatInputInteraction
+  ): Promise<APIMessage | Message<boolean> | undefined> {
     const { client } = container;
     const initialMessageFromBot: SapphireMessageRequest = await interaction.reply({
       content: this.messageWhenExecutingCommand,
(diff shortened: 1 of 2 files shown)


src/codeyCommand.ts
```diff
@@ -47,13 +47,11 @@ export class CodeyCommand extends SapphireCommand {
...
-  public async messageRun(message: Message): Promise<Message<boolean>> {
+  public async messageRun(message: Message): Promise<Message<boolean> | undefined> {
     const { client } = container;
-    const initialMessageFromBot: SapphireMessageRequest = await message.channel.send({
-      content: this.messageWhenExecutingCommand
-    });
     try {
-      const successResponse = await this.executeCommand(client, message, initialMessageFromBot);
+      const successResponse = await this.executeCommand(client, message);
+      if (!successResponse) return;
```
marko-polo-cheno, line 54: if I do return here, what happens?
can we just throw instead for err msg?

src/codeyCommand.ts
```diff
@@ -67,7 +65,9 @@ export class CodeyCommand extends SapphireCommand {
   }
 
   // Slash command
-  public async chatInputRun(interaction: SapphireCommand.ChatInputInteraction): Promise<APIMessage | Message<boolean>> {
+  public async chatInputRun(
+    interaction: SapphireCommand.ChatInputInteraction
+  ): Promise<APIMessage | Message<boolean> | undefined> {
```
marko-polo-cheno, line 70: not sure what undefined does, i dont love it

marko-polo-cheno (changes requested):

src/codeyCommand.ts
marko-polo-cheno: ig my question is, what is it for?

src/codeyCommand.ts
Picowchew: For example, in the `ping` command, the edit of the `Ping?` message already occurs inside `PingCommand`, so nothing needs to be sent from the abstraction of `CodeyCommand`. This is signified through `return Promise.resolve('');`, so `!successResponse` evaluates to true, effectively returning from the function and doing nothing more. This is the expected behaviour, so no error is thrown.

src/commands/ping/ping.ts
```diff
@@ -1,24 +1,43 @@
...
-  const content = `Pong from JavaScript! Bot Latency ${botLatency}ms. API Latency ${apiLatency}ms.`;
-  return new Promise((resolve, _reject) => resolve(content));
+  if (initialMessageFromBot) {
+    const apiLatency = getApiLatency(initialMessageFromBot, messageFromUserAsMessage);
+    return new Promise((resolve, _reject) => resolve(content(botLatency, apiLatency)));
+  }
+  const initialPing = await messageFromUserAsMessage.channel.send(initialPingContent);
+  const apiLatency = getApiLatency(initialPing, messageFromUserAsMessage);
+  initialPing.edit(content(botLatency, apiLatency));
+  return Promise.resolve('');
```
marko-polo-cheno, line 36: if we always return smth, will we ever get fail msg?

marko-polo-cheno: maybe i just dont get it 😓

Picowchew: Yup, if any line of code before the return does not work, then the failure message will send.

alexz429 (approved):

marko-polo-cheno (approved):

PR #237, opened 2022-07-14
Adding options and subcommands to CodeyCommand

mcpenguin:
# Summary of Changes
- Added the options and subcommands functionalities into the CodeyCommand class
- Slashified the rollDice command (which requires options)
- Fixed a bug where slash commands would not be updated if they had already been registered

### Options

**Regular command**
image

**Slash command**
image
image

### Sub-Commands

**Regular command**
image

**Slash command**
image

After using `/ping one`:
image

After using `/ping two`:
image

## Notes
- The slash command parameter is type-enforced and required. In other words, you need to put in a sides parameter in order to run the command, and said parameter must be an integer.
- The `ping one` and `ping two` commands are only placeholders to verify that simple subcommands work. My next step will be to slashify coin - when I have done this, I will replace the ping back with the "original" command and show the coin screenshots instead.

### DIFF
--- a/src/codeyCommand.ts
+++ b/src/codeyCommand.ts
@@ -232,9 +232,13 @@ export class CodeyCommand extends SapphireCommand {
       );
     }
 
+    let subcommandName = message.content.split(' ')[1];
+    /** The command details object to use */
+    let commandDetails = this.details.subcommandDetails[subcommandName] ?? this.details;
+
     try {
-      const successResponse = await this.details.executeCommand(client, message, initialMessageFromBot, args);
-      switch (this.details.codeyCommandResponseType) {
+      const successResponse = await commandDetails.executeCommand(client, message, initialMessageFromBot, args);
+      switch (commandDetails.codeyCommandResponseType) {
         case CodeyCommandResponseType.EMBED:
           return await message.channel.send({ embeds: [<MessageEmbed>successResponse] });
         case CodeyCommandResponseType.STRING:
@@ -242,7 +246,7 @@ export class CodeyCommand extends SapphireCommand {
       }
     } catch (e) {
       console.log(e);
-      return await message.channel.send(this.details.messageIfFailure);
+      return await message.channel.send(commandDetails.messageIfFailure);
     }
   }
 
@@ -262,11 +266,14 @@ export class CodeyCommand extends SapphireCommand {
         .map((commandOption) => commandOption.name)
         .map((commandOptionName) => ({ [commandOptionName]: interaction.options.get(commandOptionName)?.value }))
     );
+    
+    /** The command details object to use */
+    let commandDetails = this.details.subcommandDetails[interaction.options.getSubcommand()] ?? this.details;
 
     if (isMessageInstance(initialMessageFromBot)) {
       try {
-        const successResponse = await this.details.executeCommand(client, interaction, initialMessageFromBot, args);
-        switch (this.details.codeyCommandResponseType) {
+        const successResponse = await commandDetails.executeCommand(client, interaction, initialMessageFromBot, args);
+        switch (commandDetails.codeyCommandResponseType) {
           case CodeyCommandResponseType.EMBED:
             const currentChannel = (await client.channels.fetch(interaction.channelId)) as TextChannel;
             return currentChannel.send({ embeds: [<MessageEmbed>successResponse] });
@@ -275,9 +282,9 @@ export class CodeyCommand extends SapphireCommand {
         }
       } catch (e) {
         console.log(e);
-        return interaction.editReply(this.details.messageIfFailure);
+        return interaction.editReply(commandDetails.messageIfFailure);
       }
     }
-    return interaction.editReply(this.details.messageIfFailure);
+    return interaction.editReply(commandDetails.messageIfFailure);
   }
 }
--- a/src/codeyCommand.ts
+++ b/src/codeyCommand.ts
@@ -272,10 +272,10 @@ export class CodeyCommand extends SapphireCommand {
     );
 
     // Get subcommand name
-    let subcommandName: string = '';
+    let subcommandName = '';
     try {
       subcommandName = interaction.options.getSubcommand();
-    } catch (e: any) {}
+    } catch (e) {}
     /** The command details object to use */
     const commandDetails = this.details.subcommandDetails[subcommandName] ?? this.details;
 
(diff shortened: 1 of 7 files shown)


mcpenguin:
Next on the list will be to do sub commands. I think some refactoring is needed first though, which I'll do next.

src/codeyCommand.ts
```diff
@@ -39,21 +150,44 @@ export class CodeyCommand extends SapphireCommand {
...
+    }
+    return builder;
+  }
 
   // Register application commands
   public override registerApplicationCommands(registry: ChatInputCommand.Registry): void {
-    registry.registerChatInputCommand((builder) => builder.setName(this.name).setDescription(this.description));
+    // This ensures any new changes are made to the slash commands are made
+    ApplicationCommandRegistries.setDefaultBehaviorWhenNotIdentical(RegisterBehavior.Overwrite);
+    registry.registerChatInputCommand(this.getSlashCommandBuilder());
```
marko-polo-cheno, line 171: oops lint

src/codeyCommand.ts
```diff
@@ -10,23 +31,113 @@ export type SapphireMessageExecuteType = (
...
-};
-
 // Type of response the Codey command will send
 export enum CodeyCommandResponseType {
   STRING,
   EMBED
 }
 
+// Command options
+export enum CodeyCommandOptionType {
```
marko-polo-cheno, line 46: would u mind commenting these big things just so ppl in future can pick up slash commands easier?

marko-polo-cheno (changes requested):

Picowchew (commented):
Great work so far! Out of curiosity, is there a reason why the naming is `options` (in `CodeyCommandDetails`) instead of `parameters` or `arguments`? I would think that the latter is more intuitive.

mcpenguin:
> Great work so far! Out of curiosity, is there a reason why the naming is `options` (in `CodeyCommandDetails`) instead of `parameters` or `arguments`? I would think that the latter is more intuitive.

In Discord.js, the parameters are listed as "options", so I chose to stick with the name to avoid confusion. I can definitely change it to parameters if needed though.

src/codeyCommand.ts
```diff
@@ -10,74 +33,253 @@ export type SapphireMessageExecuteType = (
...
+      content: this.details.messageWhenExecutingCommand
     });
+
+    // Get command arguments
+    const args: CodeyCommandArguments = {};
+    for (const commandOption of this.details.options!) {
+      args[commandOption.name] = <CodeyCommandArgumentValueType>(
+        await commandArgs.pick(<keyof ArgType>commandOption.type)
+      );
+    }
```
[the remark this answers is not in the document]
mcpenguin: Honestly, this is why regular commands kind of suck, lol. If we use this, we are restricted to only having commands that have only one parameter that can span multiple words.

With our old framework, the following command would not be possible:

`/command  [option1: This option spans multiple words!] [option2: This option also spans multiple words!]`

whereas this is supported with slash commands.

src/codeyCommand.ts
```diff
@@ -10,72 +33,256 @@ export type SapphireMessageExecuteType = (
   client: SapphireClient<boolean>,
   // Message is for normal commands, ChatInputInteraction is for slash commands
   messageFromUser: Message | SapphireCommand.ChatInputInteraction,
-  initialMessageFromBot?: SapphireMessageRequest
+  initialMessageFromBot?: SapphireMessageRequest,
+  // Command arguments
+  args?: CodeyCommandArguments
```
Picowchew, line 38: Without any arguments, it looks like `args` is still defined to be `{}` and then passed into `commandDetails.executeCommand`, so I think `args` can be required instead. Since required parameters must come before optional parameters, I guess `initialMessageFromBot` and `args` should be switched around.

src/codeyCommand.ts
```diff
@@ -10,72 +33,256 @@ export type SapphireMessageExecuteType = (
...
+      );
+    }
+
+    const subcommandName = message.content.split(' ')[1];
+    /** The command details object to use */
+    const commandDetails = this.details.subcommandDetails[subcommandName] ?? this.details;
+
     try {
-      const successResponse = await this.executeCommand(client, message);
+      const successResponse = await commandDetails.executeCommand(client, message, undefined, args);
```
Picowchew, line 243: With `initialMessageFromBot` and `args` switched around, I guess the code will look cleaner in any case, since you can just leave out the `undefined` and do `const successResponse = await commandDetails.executeCommand(client, message, args);`.

Picowchew (commented):
I see, alright in that case, it makes sense to leave the naming as `options`.

src/codeyCommand.ts
```diff
@@ -10,72 +33,256 @@ export type SapphireMessageExecuteType = (
...
+  builder: SlashCommandBuilder,
+  subcommandDetails: CodeyCommandDetails
+): SlashCommandSubcommandsOnlyBuilder => {
+  const subcommandBuilder = new SlashCommandSubcommandBuilder();
+  subcommandBuilder.setName(subcommandDetails.name);
+  subcommandBuilder.setDescription(subcommandDetails.description);
+  for (const commandOption of subcommandDetails.options) {
+    setCommandOption(subcommandBuilder, commandOption);
+  }
+  return builder.addSubcommand(subcommandBuilder);
```
alexz429, line 195: what does making a subcommand do? I see you're still manually checking for subcommands later on in the code.

alexz429 (approved):
LGTM, asked some questions abt the new framework but all good

src/codeyCommand.ts
```diff
@@ -10,72 +33,256 @@ export type SapphireMessageExecuteType = (
   client: SapphireClient<boolean>,
   // Message is for normal commands, ChatInputInteraction is for slash commands
   messageFromUser: Message | SapphireCommand.ChatInputInteraction,
+  // Command arguments
+  args: CodeyCommandArguments,
```
marko-polo-cheno, line 37: ever null?

src/codeyCommand.ts
```diff
@@ -10,72 +33,256 @@ export type SapphireMessageExecuteType = (
...
+    }
+    return builder;
+  }
 
   // Register application commands
   public override registerApplicationCommands(registry: ChatInputCommand.Registry): void {
-    registry.registerChatInputCommand((builder) => builder.setName(this.name).setDescription(this.description));
+    // This ensures any new changes are made to the slash commands are made
+    ApplicationCommandRegistries.setDefaultBehaviorWhenNotIdentical(RegisterBehavior.Overwrite);
+    // We need to do this because TS is weird
```
marko-polo-cheno, line 220: all of this is kinda weird XD

marko-polo-cheno (approved):
big one
i hope nothing breaks lol

src/codeyCommand.ts
mcpenguin: Nope, it can be empty

PR #244, opened 2022-07-16
enable profile about to default to self profile

alexz429:
`.profile` has always felt more intuitive if it shows the caller's profile by default, this issue implements that.
image


marko-polo-cheno (approved):
so big brain

PR #249, opened 2022-07-19
Small fixes to slash commands, refactored flipcoin/rolldice

mcpenguin:
Made some small fixes to bugs I found after merging the previous slash commands PR.

Also, TS was complaining that the commands "flipCoin" and "rollDice" should be named lower-case, not camel case, so I've made this change accordingly.


Picowchew (commented):
What exactly was the error that occurred when the commands were in camel case?

marko-polo-cheno (approved):

PR #250, opened 2022-07-21
setup local dev and staging to use yarn

alexz429:
Change dockerfiles for local and staging environments to use yarn instead of node for consistency.

Update dockerfiles to use alpine 18 to prevent compatibility issues.

Once this is pushed, the same dockerfile should work for production with no extra problem.

Dockerfile tested to work in local production, waiting on staging to verify.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,13 @@
-FROM node:16-alpine
+FROM node:18-alpine
 
 # Create app directory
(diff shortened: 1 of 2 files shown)


marko-polo-cheno (approved):
now, my question is, how to QA this XD

Picowchew (approved):

PR #257, opened 2022-07-29
add daily lazy loading to github

### LINKED ISSUE #255: [FEATURE] Add database caching for app version
Currently, the GitHub API, which we use to retrieve the app version, has a rate limit. We should implement some sort of database caching of the app version to limit the number of calls to the API.

alexz429:
resolves #255 

Caches github info and release daily to prevent rate limiting.


mcpenguin (approved):
LGTM!

PR #258, opened 2022-07-30
Slashify coin

mcpenguin:
# Summary of Changes
Implemented the CodeyCommand framework for Codey Coin, adding regular and slash commands for coin.

## Todos
- Maybe make the username mentionables?
- Make codeycoin leaderboard (this could be a separate PR)

# Screenshots

## Slash Command Menu
image

### Regular command
image

### Slash command
image
image
image

### Regular command
image

### Slash command
image
image

### Regular command
image
image

### Slash command
image
image
image

### Regular command
image

### Slash command
image
image

### Regular command
image

### Slash command
image
image
image


src/commands/coin/coin.ts
```diff
@@ -24,119 +290,126 @@ import { EMBED_COLOUR } from '../../utils/embeds';
...
+//       amount,
+//       UserCoinEvent.AdminCoinUpdate,
+//       reason ? reason : null,
+//       message.author.id
+//     );
+//     // Get new balance
+//     const newBalance = await getCoinBalanceByUserId(user.id);
+//     return message.reply(`${user.username} now has ${newBalance} Codey coins 🪙.`);
+//   }
+// }
```
Picowchew, line 415: Could we just delete the commented code?

Picowchew (changes requested):
It looks like in a few places, `commandDetails.messageIfFailure` is not sent anymore. Do we want to keep this so the user is aware that something is not working? In addition, in `coin.ts`, there appears to be error catching, in which the error is then logged or nothing is done at all. In these cases, do we also want the user to be aware that there is an error?

mcpenguin:
> It looks like in a few places, commandDetails.messageIfFailure is not sent anymore. Do we want to keep this so the user is aware that something is not working? In addition, in coin.ts, there appears to be error catching, in which the error is then logged or nothing is done at all. In these cases, do we also want the user to be aware that there is an error?
Good catch, I've reimplemented this. I removed the messages at first for testing purposes, and never put them back.

src/codeyCommand.ts
```diff
@@ -292,11 +308,12 @@ export class CodeyCommand extends SapphireCommand {
...
             return interaction.editReply(<string>successResponse);
         }
-      } catch (e) {
-        console.log(e);
-        return interaction.editReply(commandDetails.messageIfFailure!);
       }
+    } catch (e) {
+      console.log(e);
+      return await interaction.editReply(
+        commandDetails.messageIfFailure ?? 'Codey backend error - contact a mod for assistance.'
```
Picowchew, line 315: I think it would be a good idea to put `'Codey backend error - contact a mod for assistance.'` into a constant since it is used more than once.

src/components/coin.ts
```diff
@@ -105,9 +113,13 @@ export const adjustCoinBalanceByUserId = async (
...
 ): Promise<void> => {
-  const oldBalance = await getCoinBalanceByUserId(userId);
-  const newBalance = Math.max(oldBalance + amount, 0);
-  await changeDbCoinBalanceByUserId(userId, oldBalance, newBalance, event, reason, adminId);
+  try {
+    const oldBalance = await getCoinBalanceByUserId(userId);
+    const newBalance = Math.max(oldBalance! + amount, 0);
+    await changeDbCoinBalanceByUserId(userId, oldBalance!, newBalance, event, reason, adminId);
+  } catch (e) {
+    console.log(e);
```
Picowchew, line 121: > In addition, in coin.ts, there appears to be error catching, in which the error is then logged or nothing is done at all. In these cases, do we also want the user to be aware that there is an error?

This is one of the few instances of what I meant by "the error is then logged."

src/components/coin.ts
```diff
@@ -122,17 +134,19 @@ export const changeDbCoinBalanceByUserId = async (
...
+      `
+      INSERT INTO user_coin (user_id, balance) VALUES (?, ?)
+      ON CONFLICT(user_id)
+      DO UPDATE SET balance = ?`,
+      userId,
+      newBalance,
+      newBalance
+    );
+    await createCoinLedgerEntry(userId, oldBalance, newBalance, event, reason, adminId);
+  } catch (e) {}
```
Picowchew, line 149: > In addition, in coin.ts, there appears to be error catching, in which the error is then logged or nothing is done at all. In these cases, do we also want the user to be aware that there is an error?

This is one of the few instances of what I meant by "nothing is done at all."

Picowchew (changes requested):

src/components/coin.ts
mcpenguin: Oh, whoops, that was from my debugging, I'll remove the try catch blocks from these

src/components/coin.ts
mcpenguin: Similar to your previous comment - I've fixed this

src/codeyCommand.ts
mcpenguin: This is now done

src/commands/coin/coin.ts
```diff
@@ -10,7 +17,266 @@ import {
...
+      name: 'reason',
+      description: 'The reason why we are adjusting the balance,',
+      type: CodeyCommandOptionType.STRING,
+      required: false
+    }
+  ],
+  subcommandDetails: {}
+};
+
+const coinInfoCommandDetails: CodeyCommandDetails = {
```
Picowchew, line 259: This should go immediately after `coinInfoExecuteCommand`?

src/components/coin.ts
```diff
@@ -165,15 +165,17 @@ export const createCoinLedgerEntry = async (
...
+    const db = await openDB();
+    await db.run(
+      `
+      INSERT INTO user_coin_bonus (user_id, bonus_type, last_granted) VALUES (?, ?, CURRENT_TIMESTAMP)
+      ON CONFLICT(user_id, bonus_type)
+      DO UPDATE SET last_granted = CURRENT_TIMESTAMP`,
+      userId,
+      bonusType
+    );
+  } catch (e) {}
```
Picowchew, line 178: Remove the try and catch blocks here as well?

src/commands/coin/coin.ts
```diff
@@ -10,7 +17,266 @@ import {
...
+  return `${user.username} now has ${newBalance} Codey coins 🪙.`;
+};
+
+const coinAdjustCommandDetails: CodeyCommandDetails = {
+  name: 'adjust',
+  aliases: ['a'],
+  description: 'Adjust the coin balance of a user.',
+  detailedDescription: `**Examples:**
+\`${container.botPrefix}coin adjust @Codey 100\`
+\`${container.botPrefix}coin adjust @Codey -100 Codey broke.\``,
```
marko-polo-cheno, line 63: what is Codey broke.

src/components/coin.ts
```diff
@@ -68,7 +68,7 @@ export interface UserCoinBonus {
   last_granted: Date;
 }
 
-export const getCoinBalanceByUserId = async (userId: string): Promise<number> => {
+export const getCoinBalanceByUserId = async (userId: string): Promise<number | undefined> => {
```
marko-polo-cheno, line 71: will we ever have undefined?

src/codeyCommand.ts
```diff
@@ -252,12 +247,14 @@ export class CodeyCommand extends SapphireCommand {
       }
     } catch (e) {
       console.log(e);
-      return await message.channel.send(commandDetails.messageIfFailure!);
+      return await message.channel.send(commandDetails.messageIfFailure ?? defaultBackendErrorMessage);
```
Picowchew, line 250: Looking back, what was wrong with the previous implementation (i.e. before `messageIfFailure? = 'Codey backend error - contact a mod for assistance';` was replaced with `messageIfFailure?: string;`)?

Picowchew (changes requested):

src/commands/coin/coin.ts
```diff
@@ -10,7 +17,266 @@ import {
...
+      • Participating in CSC events
+      • Being active on Discord
+      ---Daily bonus - your first message of the day on CSC's Discord will grant some Codey coins
+      ---Activity bonus - your first message of every 5 minutes on CSC's Discord will grant some Codey coins`
+    },
+    {
+      name: '🪙   What Can I Do With Codey Coins?',
+      value: `Use Codey coins to:
+      • Play Casino games such as Blackjack
+      • Buy CSC Swag (more info to come!)`
```
marko-polo-cheno, line 179: change this to Buy in-server CSC swag

src/commands/coin/coin.ts
```diff
@@ -1,7 +1,14 @@
-import { ApplyOptions } from '@sapphire/decorators';
-import { Args, container } from '@sapphire/framework';
-import { SubCommandPluginCommand, SubCommandPluginCommandOptions } from '@sapphire/plugin-subcommands';
-import { Message, MessageEmbed } from 'discord.js';
+import { Command, container } from '@sapphire/framework';
```
marko-polo-cheno, line 1: lol can u rearrange code so it isnt 300 lines of green then 100 lines of red 😓 makes pr reviews easier

src/codeyCommand.ts
mcpenguin: The compiler was giving errors when I tried the previous implementation, so I figured I would use this instead.

src/components/coin.ts
marko-polo-cheno: not sure why try catch is needed
if needed, should log smth

marko-polo-cheno (changes requested):

src/commands/coin/coin.ts
mcpenguin: That's the reason. Although at the moment I don't think this will work, because all parameters need to be "one word"

src/components/coin.ts
mcpenguin: We won't, we just did because I had try catches that I removed earlier

src/commands/coin/coin.ts
mcpenguin: I'm not sure what you mean by this

src/components/coin.ts
mcpenguin: Removed, this was from debugging that I forgot to remove, oops

src/codeyCommand.ts
Picowchew: What about keeping the previous implementation (for the most part) and replacing
```ts
const pingCommandDetails: CodeyCommandDetails = {
  name: 'ping',
  aliases: ['pong'],
  description: 'Ping the bot to see if it is alive.',
  detailedDescription: `**Examples:**
    \`${container.botPrefix}ping\`
    \`${container.botPrefix}pong\``,

  isCommandResponseEphemeral: true,
  messageWhenExecutingCommand: 'Ping?',
  messageIfFailure: 'Failed to receive ping.',
  executeCommand: executeCommand,
  codeyCommandResponseType: CodeyCommandResponseType.STRING,

  options: [],
  subcommandDetails: {}
};
```
with
```ts
const pingCommandDetails = new CodeyCommandDetails();
pingCommandDetails.name = 'ping';
pingCommandDetails.aliases = ['pong'];
pingCommandDetails.description = 'Ping the bot to see if it is alive.';
pingCommandDetails.detailedDescription = `**Examples:**
\`${container.botPrefix}ping\`
\`${container.botPrefix}pong\``;
pingCommandDetails.messageWhenExecutingCommand = 'Ping?';
pingCommandDetails.messageIfFailure = 'Failed to receive ping.';
pingCommandDetails.executeCommand = executeCommand;
```
?
This would lead to the default values of `CodeyCommandDetails` being used. Also, in this way, I do not think that the fields with default values need to be optional anymore? If you agree, then similar changes can be done for the command details of the other commands.

src/commands/coin/coin.ts
```diff
@@ -10,7 +17,266 @@ import {
...
+
+const coinAdjustCommandDetails: CodeyCommandDetails = {
+  name: 'adjust',
+  aliases: ['a'],
+  description: 'Adjust the coin balance of a user.',
+  detailedDescription: `**Examples:**
+\`${container.botPrefix}coin adjust @Codey 100\`
+\`${container.botPrefix}coin adjust @Codey -100 Codey broke.\``,
+
+  isCommandResponseEphemeral: true,
```
Picowchew, line 65: I think this should be false, especially because this command can only be used by administrators. so there should be a visible record of the command usage.

src/commands/coin/coin.ts
```diff
@@ -10,7 +17,266 @@ import {
...
+};
+
+const coinUpdateCommandDetails: CodeyCommandDetails = {
+  name: 'update',
+  aliases: ['u'],
+  description: 'Update the coin balance of a user.',
+  detailedDescription: `**Examples:**
+\`${container.botPrefix}coin update @Codey 100\``,
+
+  isCommandResponseEphemeral: true,
```
Picowchew, line 249: For a similar reasoning as above, I think this should be false as well.

Picowchew (changes requested):

src/commands/coin/coin.ts
mcpenguin: I prefer the first implementation, unless others feel more strongly to the second implementation.

mcpenguin:
Also, I'll just mention that because of the way Sapphire works, we can only have nesting of at most 2 in the folder structure of the commands. This is why all the subcommands for coin has to go in one file. If someone has a way of making it so that each subcommand can be in its own file under a `subcommands` folder in `commands/coin` that would be appreciated.

src/commands/coin/coin.ts
mcpenguin: (meant to reply to the first comment)

src/codeyCommand.ts
mcpenguin: Personally, I prefer the first implementation. I think it's clearer and more compact. Also, switching to the second implementation would be a lot of refactoring as well.

src/commands/coin/coin.ts
mcpenguin: Sounds good, I've added this

src/commands/coin/coin.ts
mcpenguin: Sounds good, I've also added this

src/commands/coin/coin.ts
marko-polo-cheno: rn i might see
X X'
Z Y'
Y Z'

gimme 
X X'
Y Y'
Z Z'
lmao probs just rearrange functions - so hard to read walls of green/red when the diff is actually 3 lines or smth

Picowchew:
For Sapphire, it looks like everything in the `commands` folder should be a command. As such, the command details of subcommands have been placed into separate files, in a new folder called `commandDetails`, by #259.

Picowchew (approved):
As we have discussed with each other elsewhere, we can leave any refactoring for another PR, since that would involve commands other than coin.

marko-polo-cheno (approved):
poke

PR #263, opened 2022-08-06
Codeycoin leaderboard

mcpenguin:
# Summary of Changes

Added a leaderboard for the top 10 CodeyCoin holders.

# Steps to Reproduce
Run `.coin leaderboard` or `/coin leaderboard`.

## Screenshots

The embed is the same regardless whether you use regular or slash commands.

image

### DIFF
--- a/src/commandDetails/coin/leaderboard.ts
+++ b/src/commandDetails/coin/leaderboard.ts
@@ -16,7 +16,6 @@ const getCurrentCoinLeaderboardEmbed = async (
   currentUserId: string,
   limit = 10
 ): Promise<MessageEmbed> => {
-  const currentLeaderboardEmbed = new MessageEmbed().setColor(EMBED_COLOUR).setTitle('Current CodeyCoin Leaderboard');
   // Initialise user's coin balance if they have not already
   const userBalance = await getCoinBalanceByUserId(currentUserId);
   const currentPosition = leaderboard.findIndex((userCoinEntry) => userCoinEntry.user_id === currentUserId) + 1;
@@ -30,16 +29,15 @@ const getCurrentCoinLeaderboardEmbed = async (
     const userCoinEntryText = `${i + 1}. ${userTag} - ${userCoinEntry.balance} 🪙.\n`;
     currentLeaderboardText += userCoinEntryText;
   }
-  currentLeaderboardEmbed.addFields(
-    {
-      name: 'Current Leaderboard',
-      value: currentLeaderboardText
-    },
-    {
-      name: 'Current Position',
-      value: `You are currently **#${currentPosition}** in the leaderboard with ${userBalance} 🪙`
-    }
-  );
+  const currentLeaderboardEmbed = new MessageEmbed()
+    .setColor(EMBED_COLOUR)
+    .setTitle('CodeyCoin Leaderboard')
+    .setDescription(currentLeaderboardText);
+
+  currentLeaderboardEmbed.addFields({
+    name: 'Your Position',
+    value: `You are currently **#${currentPosition}** in the leaderboard with ${userBalance} 🪙`
+  });
 
   return currentLeaderboardEmbed;
 };
--- a/src/commandDetails/coin/leaderboard.ts
+++ b/src/commandDetails/coin/leaderboard.ts
@@ -1,0 +1,62 @@
+import { container, SapphireClient } from '@sapphire/framework';
+import { Client, MessageEmbed } from 'discord.js';
+import {
+  CodeyCommandDetails,
+  CodeyCommandResponseType,
+  getUserIdFromMessage,
+  SapphireMessageExecuteType,
+  SapphireMessageResponse
+} from '../../codeyCommand';
+import { getCurrentCoinLeaderboard, UserCoinEntry } from '../../components/coin';
+import { EMBED_COLOUR } from '../../utils/embeds';
+
+const getCurrentCoinLeaderboardEmbed = async (
+  client: SapphireClient<boolean>,
+  leaderboard: UserCoinEntry[],
+  currentUserId: string,
+  limit = 10
+): Promise<MessageEmbed> => {
+  const currentLeaderboardEmbed = new MessageEmbed().setColor(EMBED_COLOUR).setTitle('Current CodeyCoin Leaderboard');
+
+  let currentLeaderboardText = '';
+  for (let i = 0; i < limit; i++) {
+    if (leaderboard.length <= i) break;
+    const userCoinEntry = leaderboard[i];
+    const user = await client.users.cache.get(userCoinEntry.user_id);
+    const userCoinEntryText = `${i + 1}. ${user?.username ?? '<unknown>'} - ${userCoinEntry.balance} 🪙.\n`;
+    currentLeaderboardText += userCoinEntryText;
+  }
+  currentLeaderboardEmbed.addFields({
+    name: 'Current Leaderboard',
+    value: currentLeaderboardText
+  });
+
+  return currentLeaderboardEmbed;
+};
+
+const coinCurrentLeaderboardExecuteCommand: SapphireMessageExecuteType = async (
+  client,
+  messageFromUser,
+  _args
+): Promise<SapphireMessageResponse> => {
+  const userid = getUserIdFromMessage(messageFromUser);
+  const leaderboard = await getCurrentCoinLeaderboard();
+  return getCurrentCoinLeaderboardEmbed(client, leaderboard, userid);
+};
+
+export const coinCurrentLeaderboardCommandDetails: CodeyCommandDetails = {
+  name: 'leaderboard',
+  aliases: ['lb'],
+  description: 'Get the current coin leaderboard.',
+  detailedDescription: `**Examples:**
+  \`${container.botPrefix}coin lb\`
+  \`${container.botPrefix}coin leaderboard\``,
+
+  isCommandResponseEphemeral: false,
+  messageWhenExecutingCommand: 'Getting the current coin leaderboard...',
+  executeCommand: coinCurrentLeaderboardExecuteCommand,
+  codeyCommandResponseType: CodeyCommandResponseType.EMBED,
+
+  options: [],
+  subcommandDetails: {}
+};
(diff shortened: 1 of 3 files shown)


src/commandDetails/coin/leaderboard.ts
```diff
@@ -0,0 +1,72 @@
...
+    if (leaderboard.length <= i) break;
+    const userCoinEntry = leaderboard[i];
+    const user = await client.users.fetch(userCoinEntry.user_id);
+    const userTag = user?.tag ?? '<unknown>';
+    const userCoinEntryText = `${i + 1}. ${userTag} - ${userCoinEntry.balance} 🪙.\n`;
+    currentLeaderboardText += userCoinEntryText;
+  }
+  currentLeaderboardEmbed.addFields(
+    {
+      name: 'Current Leaderboard',
```
Picowchew, line 35: Can we remove this heading (I think it might seem a bit redundant)? I think the contents of the leaderboard can probably go into the embed description.

src/commandDetails/coin/leaderboard.ts
```diff
@@ -0,0 +1,72 @@
...
+      value: currentLeaderboardText
+    },
+    {
+      name: 'Current Position',
```
Picowchew, line 39: I think this can be renamed to "Your Position" instead.

src/commandDetails/coin/leaderboard.ts
```diff
@@ -0,0 +1,72 @@
...
+import { getCurrentCoinLeaderboard, getCoinBalanceByUserId, UserCoinEntry } from '../../components/coin';
+import { EMBED_COLOUR } from '../../utils/embeds';
+
+const getCurrentCoinLeaderboardEmbed = async (
+  client: SapphireClient<boolean>,
+  leaderboard: UserCoinEntry[],
+  currentUserId: string,
+  limit = 10
+): Promise<MessageEmbed> => {
+  const currentLeaderboardEmbed = new MessageEmbed().setColor(EMBED_COLOUR).setTitle('Current CodeyCoin Leaderboard');
```
Picowchew, line 19: I think "Current" can be removed from the title, so it would just be "CodeyCoin Leaderboard" for conciseness.

Picowchew (changes requested):
I do not think that bots should be able to have Codey coins, and therefore bots should definitely not appear on the leaderboard. While this change might seem to be out of scope of this PR, I think it would be nice to include this change since the leaderboard is client-facing.

Also, it looks like there are some linter warnings.

src/commandDetails/coin/leaderboard.ts
```diff
@@ -0,0 +1,72 @@
...
+  SapphireMessageResponse
+} from '../../codeyCommand';
+import { getCurrentCoinLeaderboard, getCoinBalanceByUserId, UserCoinEntry } from '../../components/coin';
+import { EMBED_COLOUR } from '../../utils/embeds';
+
+const getCurrentCoinLeaderboardEmbed = async (
+  client: SapphireClient<boolean>,
+  leaderboard: UserCoinEntry[],
+  currentUserId: string,
+  limit = 10
```
marko-polo-cheno, line 17: maybe a const

src/components/coin.ts
```diff
@@ -135,6 +140,21 @@ export const changeDbCoinBalanceByUserId = async (
...
+/*
+  Get the leaderboard for the current coin amounts.
+*/
+export const getCurrentCoinLeaderboard = async (): Promise<UserCoinEntry[]> => {
+  const db = await openDB();
+  const res = await db.all(
+    `
+      SELECT user_id, balance
+      FROM user_coin
+      ORDER BY balance DESC
```
marko-polo-cheno, line 152: limit by same const

marko-polo-cheno (changes requested):

marko-polo-cheno:
while u are here
change coin image into codey coin emoji

src/components/coin.ts
mcpenguin: We need to get all because we get the current position of the user as well

src/commandDetails/coin/leaderboard.ts
mcpenguin: Do you think we could make this a parameter instead? Like in the command

src/commandDetails/coin/leaderboard.ts
mcpenguin: Alright this is now done

src/commandDetails/coin/leaderboard.ts
mcpenguin: Done as well

src/commandDetails/coin/leaderboard.ts
mcpenguin: I've implemented this change

src/commandDetails/coin/leaderboard.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  return currentLeaderboardEmbed;
+};
+
+const coinCurrentLeaderboardExecuteCommand: SapphireMessageExecuteType = async (
+  client,
+  messageFromUser,
+  _args
+): Promise<SapphireMessageResponse> => {
+  const userid = getUserIdFromMessage(messageFromUser);
```
Picowchew, line 50: `userid` -> `userId`?

src/commandDetails/coin/leaderboard.ts
```diff
@@ -0,0 +1,70 @@
...
+  return getCurrentCoinLeaderboardEmbed(client, leaderboard, userid);
+};
+
+export const coinCurrentLeaderboardCommandDetails: CodeyCommandDetails = {
+  name: 'leaderboard',
+  aliases: ['lb'],
+  description: 'Get the current coin leaderboard.',
+  detailedDescription: `**Examples:**
+  \`${container.botPrefix}coin lb\`
+  \`${container.botPrefix}coin leaderboard\``,
```
Picowchew, line 61: There should not be spaces at the beginning of each example because they would show in the resulting string.

Picowchew (changes requested):
With regard to my previous comment about bots, I made a new issue, #265.

src/commandDetails/coin/leaderboard.ts
marko-polo-cheno: sure, probably cap at 150 tho, I don't want severe performance issues with ppl spamming .lb 2000

src/components/coin.ts
marko-polo-cheno: oh, hmmmmmm... maybe a count of all-greater-than with a lookup of this user's coin would be the best performance thingy

mcpenguin: Yeah I think that would be good, I'll implement that change

src/commandDetails/coin/leaderboard.ts
mcpenguin: I might cap it at 10 for now, since embeds have a max size. We can create an issue to introduce pagination to the leaderboards if needed to account for larger constants.

mcpenguin: For now the limit will be set to 10. But I'll keep the limit a parameter so that later on the leaderboard size can be increased.

src/components/coin.ts
```diff
@@ -135,6 +140,37 @@ export const changeDbCoinBalanceByUserId = async (
...
+  Get the leaderboard for the current coin amounts.
+*/
+export const getCurrentCoinLeaderboard = async (limit = 10): Promise<UserCoinEntry[]> => {
+  const db = await openDB();
+  const res = await db.all(
+    `
+      SELECT user_id, balance
+      FROM user_coin
+      ORDER BY balance DESC
+      LIMIT ${limit}
```
marko-polo-cheno, line 153: In the past I used "?"
instead of ${}
is there a big difference? Just curious

src/components/coin.ts
```diff
@@ -135,6 +140,37 @@ export const changeDbCoinBalanceByUserId = async (
...
+/*
+  Get the position for a user id's balance
+*/
+export const getUserIdCurrentCoinPosition = async (userId: string): Promise<number> => {
+  const db = await openDB();
+  const res = await db.get(
+    `
+      SELECT COUNT(user_id) AS count
+      FROM user_coin
+      WHERE balance >= (SELECT balance FROM user_coin WHERE user_id = '${userId}')
```
marko-polo-cheno, line 168: I think > is better than >=

marko-polo-cheno, line 168: could reuse getCoinBalanceByUserId

src/components/coin.ts
```diff
@@ -61,6 +61,11 @@ export const coinBonusMap = new Map<BonusType, Bonus>([
   ]
 ]);
 
+export interface UserCoinEntry {
```
marko-polo-cheno, line 64: I can't ctrl f on phone but can u just lmk where this is used XD

src/commandDetails/coin/leaderboard.ts
```diff
@@ -0,0 +1,75 @@
...
+  UserCoinEntry,
+  getUserIdCurrentCoinPosition
+} from '../../components/coin';
+import { EMBED_COLOUR } from '../../utils/embeds';
+
+const getCurrentCoinLeaderboardEmbed = async (
+  client: SapphireClient<boolean>,
+  leaderboard: UserCoinEntry[],
+  currentUserId: string,
+  limit = 10
```
marko-polo-cheno, line 22: still prefer a constant at the top of file, so limit = CONSTANT instead of limit = magic number

marko-polo-cheno (approved):
high demand feature :) tyty
made some tiny notes

src/components/coin.ts
Picowchew: Yup, it is recommended to use "?" to avoid SQL injections.

Picowchew (approved):

src/components/coin.ts
mcpenguin: Nope, you guys are right, I'll change it to use "?"

src/components/coin.ts
mcpenguin: In the leaderboard :))

PR #269, opened 2022-08-08
[FEATURE] Add possibility of using percentage in the blackjack command

botatooo:
# Summary of Changes
For the blackjack command (`.bj`), adds the possibility of using a percentage of your total Codey coins instead of a fixed amount

# Steps to Reproduce
`.bj 10%`
image
image 2


src/commands/games/blackjack.ts
```diff
@@ -186,10 +186,21 @@ export class BlackjackCommand extends Command {
   async messageRun(message: Message, args: Args): Promise<Message> {
     // if there are no arguments, then resolve to the default bet amount; if there is only one argument and it is an
     // integer, then this is the bet amount; otherwise, reply that a valid bet amount must be entered
-    const bet = args.finished
+    const betAmount = await args.restResult<number>('integer');
+    const betPercent = await args.restResult<string>('string');
+    let bet = args.finished
```
marko-polo-cheno, line 191: where's the please enter valid bet amt?


screenshots of slash and dot command working?
suppose i did .bj -50% or 200% or 0.001%
i don't do QA

marko-polo-cheno (changes requested):

src/commands/games/blackjack.ts
botatooo: I didn't mess with the bet amount validator, so in order what happens is:
- gets percentage
- gets user's coins
- process percentage
- assign it back to the `bet` variable

and the rest of the process would go as planned

marko-polo-cheno (approved):
poggers

Picowchew (changes requested):
Thanks for your contribution! :)
I just made some comments for what appears to be a few logic flaws.

botatooo:
Thanks for pointing out those issues! I'll be addressing them tomorrow.

src/commands/games/blackjack.ts
```diff
@@ -186,12 +186,18 @@ export class BlackjackCommand extends Command {
...
 
     const { author, channel } = message;
+    const playerBalance = await getCoinBalanceByUserId(author.id);
+
+    let bet: number;
+    if (betString.includes('%')) {
+      const percentage = parseInt(betString.replace('%', '').trim(), 10) / 100;
+      bet = Math.round(playerBalance * percentage);
+    } else {
+      bet = args.finished ? DEFAULT_BET : parseInt(betString, 10);
```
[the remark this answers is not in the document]
botatooo: I've fixed this by instead implementing the default bet value in a catcher.

src/commands/games/blackjack.ts
```diff
@@ -186,12 +186,18 @@ export class BlackjackCommand extends Command {
   async messageRun(message: Message, args: Args): Promise<Message> {
     // if there are no arguments, then resolve to the default bet amount; if there is only one argument and it is an
     // integer, then this is the bet amount; otherwise, reply that a valid bet amount must be entered
-    const bet = args.finished
-      ? DEFAULT_BET
-      : await args.rest('integer').catch(() => 'please enter a valid bet amount.');
-    if (typeof bet === 'string') return message.reply(bet);
+    const betString = await args.rest('string').catch(() => 'please enter a valid bet amount.');
```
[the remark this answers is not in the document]
botatooo: I've fixed this by checking the final value using `isNaN`.

src/commands/games/blackjack.ts
```diff
@@ -206,8 +210,19 @@ export class BlackjackCommand extends Command {
...
-      return message.reply(`you don't have enough coins to place that bet. ${getEmojiByName('codeySad')}`);
+
+    if (playerBalance! < bet) {
+      if (betString.endsWith('%')) {
+        const percentNeeded = playerBalance / (playerBalance / 10);
+        return message.reply(
+          `you don't have enough coins to place that bet. (you'd need at least ${percentNeeded}%) ${getEmojiByName(
+            'codeySad'
+          )}`
+        );
```
Picowchew, line 223: Could we cap the % at 100% instead? In other words, if the player types a value greater than 100%, then the bot will reply that the maximum bet is 100%.

Picowchew (changes requested):
Thanks for the quick fixes! I left another comment. It would also be nice if you could update the screenshots. :)

Picowchew:
I don't currently have my dev environment with me, so just reading through the code more carefully, does the blackjack game even run when a percentage is given? `validateBetAmount` will return a non-empty string, and then the bot will immediately reply with `validateRes`?

src/commands/games/blackjack.ts
botatooo: Done. See line 203.

hydrobeam:
Maybe it'd be cleaner to just only have regular numbers as an option for betting? If somebody wants to bet a percentage of their balance then they can pretty easily get that info with a '/coin check', so I don't think there's that much value in having it built in

src/commands/games/blackjack.ts
Picowchew: I see that you added this check. If I am not mistaken, this `if` block can be removed? You would only not have enough coins to place a bet if it is over 100%. In addition, it looks like `percentNeeded` always evaluates to 10, which does not look right?

Picowchew:
The change is only taking place in 1 file, which doesn't look to be all too messy. If it's a feature that's going to be used, then I don't think it hurts to add it.

hydrobeam:
When blackjack is slashified it might make receiving input a little more inconvenient since we won't be able to use a NumberOption (which can set client enforceable min/max values) since the input'll have to be a string to accommodate this feature

Picowchew:
I see. I do agree that it'll be more inconvenient, but in that case, string should be fine.

botatooo:
> When blackjack is slashified it might make receiving input a little more inconvenient since we won't be able to use a NumberOption (which can set client enforceable min/max values) since the input'll have to be a string to accommodate this feature

We could add two, a number and a string option, both set to optional.

hydrobeam:
That's not very clean since then you'd always need to go through a pseudo-prompt to select one of the two because they're optional, whereas a mandatory argument is easier to reach

src/commands/games/blackjack.ts
```diff
@@ -186,23 +188,46 @@ export class BlackjackCommand extends Command {
...
       return message.reply(validateRes);
     }
 
     // check player balance and see if it can cover the bet amount
-    const playerBalance = await getCoinBalanceByUserId(author.id);
-    if (playerBalance! < bet)
-      return message.reply(`you don't have enough coins to place that bet. ${getEmojiByName('codeySad')}`);
+
+    if (playerBalance! < bet) {
+      if (betString.endsWith('%')) {
```
gzcharleszhang, line 220: we're repeating this logic multiple times, could we abstract this to a helper?

src/commands/games/blackjack.ts
```diff
@@ -186,23 +188,46 @@ export class BlackjackCommand extends Command {
...
+      if (percentage > 1) {
+        return message.reply(`You can only bet up to 100%. ${getEmojiByName('codeySad')}`);
+      }
+      bet = Math.round(playerBalance * percentage);
+    } else {
+      bet = parseInt(betString, 10);
+    }
 
-    const validateRes = validateBetAmount(bet);
+    const validateRes = validateBetAmount(bet, betString.endsWith('%'));
```
gzcharleszhang, line 211: repeated logic

src/commands/games/blackjack.ts
gzcharleszhang: I agree, let's just remove this if block. This is not necessary in terms of code logic or user experience (i.e. I'd rather know how many more coins I need to place that bet, instead of what % more I need).

src/commands/games/blackjack.ts
```diff
@@ -186,23 +188,46 @@ export class BlackjackCommand extends Command {
...
-    const bet = args.finished
-      ? DEFAULT_BET
-      : await args.rest('integer').catch(() => 'please enter a valid bet amount.');
-    if (typeof bet === 'string') return message.reply(bet);
+    const betString = await args.pick('string').catch(() => String(DEFAULT_BET));
 
     const { author, channel } = message;
+    const playerBalance = await getCoinBalanceByUserId(author.id);
+
+    if (isNaN(parseInt(betString, 10))) {
```
gzcharleszhang, line 196: Can we just `parseInt` once, and then reuse this variable? I know the unhandled `parseInt`s in later parts of this function shouldn't return `NaN`, but it's not a good practice to leave them like that.

gzcharleszhang (changes requested):

src/commands/games/blackjack.ts
```diff
@@ -186,23 +188,46 @@ export class BlackjackCommand extends Command {
...
+      return message.reply('please enter a valid bet amount.');
+    }
+
+    let bet: number;
+    if (betString.endsWith('%')) {
+      const percentage = parseInt(betString, 10) / 100;
+      if (percentage > 1) {
```
gzcharleszhang, line 203: Is there a reason why we add this logic here? We already have logic that verifies a bet cannot be greater than a player's balance.

src/commands/games/blackjack.ts
botatooo: Good point, I removed it

src/commands/games/blackjack.ts
botatooo: forgot about that, removed

src/commands/games/blackjack.ts
botatooo: you've got a point, done

mcpenguin (changes requested):
When I try to run the command it seems that I get some errors:
```
src/commands/games/blackjack.ts(233,15): error TS2448: Block-scoped variable 'playerBalance' used before its declaration.
src/commands/games/blackjack.ts(233,15): error TS2454: Variable 'playerBalance' is used before being assigned.
src/commands/games/blackjack.ts(240,24): error TS2448: Block-scoped variable 'playerBalance' used before its declaration.
src/commands/games/blackjack.ts(240,24): error TS2454: Variable 'playerBalance' is used before being assigned.

```

PR #270, opened 2022-08-09
Convert resume pdfs to images

mcpenguin:
# Summary of Changes
PDFs sent to the #resumes channel will be converted into images automatically by Codey.

## TODOs Before Merging
- We need to update the staging and production dockerfiles to include these lines that download deps necessary for the PDF-\>image converter:
```
RUN apk add --update ghostscript
RUN apk add --update graphicsmagick
```
- We need to update the staging and production configs to account for the new config variable, `RESUME_CHANNEL_ID` (see below)

# Testing Notes
1. Add the `RESUME_CHANNEL_ID` variable to your `config.json`; this is the channel where the resumes are sent (in this case it will be #resumes)
2. Send a message without any attachment: nothing should happen (this is similar to when people post google drive links to their resumes)
3. Send a message with a pdf of a resume: Codey should automatically send back an image corresponding to the resume.

## Screenshots

Nothing happens if no attachment is sent, so the screenshot for the case where a resume pdf is attached is shown below:

image


alexz429 (changes requested):
Some small optimizations for dockerfiles:
Add the run steps for mkdir and library installations at the very start, since those actions can probably be cached before any file changes.
Add /tmp to dockerignore to prevent local tmp files being sent over.

Also don't forget to add the new env var into staging and prod configs.

alexz429 (approved):

PR #272, opened 2022-08-10
fix github loading

alexz429:
Correctly activates lazy loading of github info after a day


mcpenguin (approved):
LGTM. thanks laura for noticing the bug

PR #280, opened 2022-08-12
Refactoring of codeyCommand and some executeCommand

mcpenguin:
# Summary of Changes
- Refactored CodeyCommand slightly; for regular commands, any commands that return an embeds are sent as a reply to the original message, not just sent to the channel
- Changed the executeCommands of commands that return embeds to be returned in a `{embeds: [<embed>]}` format

no screenshots attached, this is just a code change


Picowchew (approved):
LGTM!

Issue #265, opened 2022-08-07
Bots (except for irc-bridge) should be ignored and no bots should appear in coin leaderboard


Picowchew:
Only real users and the irc-bridge bot (not any other bots) should not be ignored by Codey's messageCreate listener. This is good practice, since in case a bot is hacked and coded to spam, then Codey won't be overloaded with reading all of the messages. The reason for irc-bridge being an exception is that CSC hosts this bot, so CSC is in control of it. In addition, it would be nice for irc-bridge to have a coin balance, since this would gauge activity level (no other bots would have a coin balance). However, no bots should be appearing in the Codey coin leaderboard (i.e. even irc-bridge should not be in the leaderboard), since the leaderboard should only be reserved for real users.

PR #281, opened 2022-08-12
Exclude bots from leaderboard

KevinBacabac:
https://github.com/uwcsc/codeybot/issues/265

I don't have a database to test this ticket


Picowchew (commented):
In terms of not having a database, shouldn't SQLite be installed due to being found in yarn.lock?

Picowchew:
I might as well add that the change would be taking place in the messageCreate listener instead of the leaderboard, since bots shouldn't have a coin balance *at all*. Bots not appearing in the leaderboard would just be one effect of this.

Picowchew:
The issue has been updated just now for clearer specifications; sorry about that!

KevinBacabac:
No problem it's ready to test

src/listeners/messageCreate.ts
```diff
@@ -112,6 +113,11 @@ export class MessageCreateListener extends Listener {
       return;
     }
```
Picowchew, line 114: This case can be removed, since it is already covered below.

Picowchew (changes requested):

Picowchew:
Also, do you mind adding 275813801792634880 as IRC_USER_ID for staging and updating README.md for the new config variable IRC_USER_ID?

Picowchew (approved):
LGTM!

PR #282, opened 2022-08-13
Make `/coin check` default to the caller

hydrobeam:
# Related Issues

N/A

# Summary of Changes

In comparison to `.coin`, doing `/coin check` and having to look for the target user was pretty time consuming when somebody just wanted to check their own balance. This PR makes the `user` argument optional and defaults to checking the balance of the user who called the command if the arg is not provided. 

- Balance not present + coin check optional
image

- .coin default behaviour checks out + message changes depending on how user is called
image

# Steps to Reproduce

`/coin check` and pass in no arguments.
OR
`.coin check` and pass in no arguments.

## Extra commentary

Not sure if `messageFromUser.member?.user` is the "right" way to get the user, since for some reason accessing `member` can return null from a `Message`?🤷


src/commandDetails/coin/check.ts
```diff
@@ -13,12 +13,11 @@ import { getCoinEmoji } from '../../components/emojis';
...
   _client,
-  _messageFromUser,
+  messageFromUser,
   args
 ): Promise<SapphireMessageResponse> => {
-  // Mandatory argument is user
-  const user = <User>args['user'];
-
+  // use the caller as a default user if no argument is provided
+  const user = <User>args['user'] ?? messageFromUser.member?.user;
```
mcpenguin, line 20: Does `messageFromUser.user` work?

mcpenguin:
Also please leave screenshots :))

src/commandDetails/coin/check.ts
hydrobeam: Nope, `user` is not a valid field for a `Message`. `author` exists for message, but I don't really know in what situation accessing member could be null for a message 🤷 so it's probably not a concern

hydrobeam:
Decided to remove the `balance` subcommand since having `/coin check` default to self makes it redundant. To maintain appearances, I added a check to see if the caller is running the method without arguments and changes the initial message to `You have` accordingly.

Also reworked `getUserIDFromMessage` to be more general

src/commandDetails/coin/check.ts
```diff
@@ -5,28 +5,39 @@ import {
...
   } catch (e) {
     return `Could not fetch the user's balance, contact a mod for help`;
   }
+
   // Show coin balance
-  return `${user.username} has ${balance} Codey coins ${getCoinEmoji()}.`;
+  return `${displayMessage} ${balance} Codey coins ${getCoinEmoji()}.`;
 };
 
 export const coinCheckCommandDetails: CodeyCommandDetails = {
```
mcpenguin, line 43: If you're removing balance, make `aliases` also have `balance, bal and b`, since people use these commands regularly.

mcpenguin (changes requested):
small changes

src/commandDetails/coin/check.ts
hydrobeam: Done

mcpenguin (approved):
LGTM! Feel free to merge after you fix the linter.

hydrobeam:
> LGTM! Feel free to merge after you fix the linter.

Fixed the linter, but don't have write access :see_no_evil:

PR #286, opened 2022-08-14
Fixed local:run script to create tmp directory

### LINKED ISSUE #285: Script for local:run does not create tmp directory
Because the script for local:run does not create the tmp directory, resume previews do not work.

Picowchew:
# Related Issues

Resolves #285 

# Summary of Changes

Added `mkdir -p tmp` to the local:run script to make the tmp directory

# Steps to Reproduce

When running the bot locally, see that the tmp directory is created if non-existent and no error is thrown if already existent. In addition, resume previews should now work.
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "",
   "main": "dist/src/bot.js",
   "scripts": {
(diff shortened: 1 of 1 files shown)


mcpenguin (approved):
LGTM!

PR #287, opened 2022-08-14
Resume staging fixes

mcpenguin:
Codey staging dies if resume pdf has > 2 pages
So we now assert resume is 1 page


Picowchew (approved):
LGTM!

PR #312, opened 2022-08-17
Migrate ready off sapphire

KevinBacabac:
## Summary of Changes
Migrates ready logic (after startup) to a separate discord.js based file.

## PRs
Resolves https://github.com/uwcsc/codeybot/issues/314

## Steps to Reproduce
Behaviour is confirmed to be the same with no regressions.


src/listeners/ready.ts
```diff
@@ -1,16 +1,9 @@
 import type { ListenerOptions, Piece, PieceContext } from '@sapphire/framework';
 import { Listener, Store } from '@sapphire/framework';
```
mcpenguin, line 2: Why are we still using Sapphire's listener?

KevinBacabac: For the record we removed this file after deciding we don't use the store feature

Picowchew (approved):
LGTM!

botatooo (approved):
LGTM

PR #304, opened 2022-08-15
feat(interviewer): Slashify and refactor interviewer commands 

hydrobeam:
## Summary of Changes

Slashified every subcommand in `src/commands/interviewer/interviewer.ts`. Following the trend in #258, moved subcommands out to the `commandDetails` directory. 

#### Choices
- Currently used by `list` under `src/commanDetails/interviewer/list.ts`
#### Components
- refer to `src/interaction-handlers/interview/domain.ts` for some insight
- Currently, the `domain` subcommand implements a `SelectMenu` component to allow multiselection/setting of domains.

## Motivation and Explanation

There's a push to make slash commands viable, plus refactors the code to make it easier to refactor.

## Related Issues

closes #283 

#### `domain`
Video showcasing `domain subcommand`
https://user-images.githubusercontent.com/70682032/185033616-160b6a33-3051-4871-96e3-6d08cbd75623.mp4

#### `list`
- optional parameter
image
- choice list
image
- working
image
- regular command with no args too

#### Remaining subcommands

image
- `signup`

image
-  `resume`

image
-  `clear` + alias works

image
- `profile`

**NOTE**: profile is a little messed up because the expected message type is an embed, and for some reason because of initialcommand shenanigans when an embed is expected it sends a separate message instead of modifying the original one.
regular commands dont't do this, only slash

image

- `pause`
image

## Steps to Reproduce

run the commands :flushed:

## Further Information and Comments
quite a pain to do + a lot of boilerplate, components pog

PR #271, opened 2022-08-10
Migrate discord.js to v14. Not ready. Awaiting sapphire

KevinBacabac:
Some code I thought I'd share. The discord.js is fully migrated now.
But there's sapphire crashes we'll need to wait for first. https://github.com/sapphiredev/framework/pull/485

PR #327, opened 2022-08-18
feat: Allow adding components to commands

hydrobeam:
## Summary of Changes

- Allowed adding components to messages. These work for both Slash and Regular commands.
- Added the type `components` to `CodeyCommandDetails` to allow sub-command declarations in `commandDetails` to declare components.
   - These are done through builders functions provided by discord.js directly, no abstraction.
   - **WARNING**: Was forced to use `discord.js v13` names, so #271 will have to be updated.
- **How it works**: If the `commandDetails` contains any components,  then don't execute the command and let the class under `interaction-handlers` do the job,
- `interaction-handlers` is currently pretty sapphire intertwined, but it shouldn't be tough to detangle imo.
- Could easily be used with Buttons, Modals, etc ...  
- See [here](https://www.sapphirejs.dev/docs/Guide/application-commands/interaction-handlers/what-are-they) for an overview of interactions (sapphire docs)

## Motivation and Explanation

Components can do cool things, like implement SelectMenus + other interactions:

https://discordjs.guide/interactions/select-menus.html#building-and-sending-select-menus

## Related Issues

N/A

## Steps to Reproduce

N/A

## Demonstration of Changes

See #304 `/interviewer domain` for a showcase of an implementation


src/codeyCommand.ts
```diff
@@ -154,13 +165,14 @@ export class CodeyCommandDetails {
...
   codeyCommandResponseType?: CodeyCommandResponseType = CodeyCommandResponseType.STRING;
-
   /** Options for the Codey command */
   options: CodeyCommandOption[] = [];
   /** Subcommands under the CodeyCommand */
   subcommandDetails: { [name: string]: CodeyCommandDetails } = {};
   /** The default subcommand to execute if no subcommand is specified */
   defaultSubcommandDetails?: CodeyCommandDetails;
+  /** Components to be shown in the message, type stolen from discord.js */
+  components?: (MessageActionRow | (Required<BaseMessageComponentOptions> & MessageActionRowOptions))[] = [];
```
mcpenguin, line 175: Extract `MessageActionRow | (Required<BaseMessageComponentOptions> & MessageActionRowOptions))` into a `CodeyComponent` type, and use that here? That way if the component class ever changes, we can just change that.

mcpenguin (changes requested):
This is a little weird, but if it works, it works

hydrobeam:
> This is a little weird, but if it works, it works


Yeah taking another look at this implementation and after having worked on #329, it doesn't really make sense for `components` to be treated any differently from embeds. So `executeCommand` should be used to generate the component, and then the actual managing of the resonse should be done in the `run` method under the exported class in `interaction-handlers`. 

My approach in #304 was to have `interaction-handlers` call back to `executeCommand`, which is weird and doesn't follow the trend established by other commands.  Would be worth introducing an optional `componentHandler` function or something like that to keep the logic centralized.

mcpenguin:
I think a `ComponentHandler`, or even a more generic `Handler` class would be good to centralize the logic, as you mentioned. In the future, as part of desapphirication of the interaction handler, we could employ the observer pattern or something similar (haha CS246 go brrrr) to register listeners for interactions for responding to components. That is outside the scope of this ticket though - I think making a handler is the right way to go here.

mcpenguin:
Closing - PR is not needed anymore

PR #232, opened 2022-07-12
Do not edit messages anymore

Picowchew:
# Related Issues

N/A

# Summary of Changes

Modified codeyCommand so that messages from non-slash commands are not edited anymore

# Steps to Reproduce

Try out non-slash commands
image

PR #329, opened 2022-08-18
feat: Do not edit slash commands after sending 

hydrobeam:
## Summary of Changes

- Removed process of sending an initial message, then subsequently editing the command for slash commands (already done for regular commands in #232 )
- Removed CodeyCommandResponseType and the accompanying field in `commandDetails`
- Removed `initialMessageFromBot`argument from `SapphireMessageExecuteType`
- Adjusted imports/fields for all relevant commands
- Refactored `ping` implementation slightly to work around this

## Motivation and Explanation

Originally (and only for slash commands), the bot would send an initial message, wait for the output, and then edit the reply (if the output was a string, or send a new message if the output was an embed). This didn't really make sense since it produces weird looking bot interactions which always had the `(edited)` tag, and also just broke with embeds, see:

- slash vs no slash

image

 - edited tag always present

image

image

This behaviour wasn't really necessary so I modified `chatInputRun` to act without it. Also, being able to cleanup uncessary types like `CodeyResponseType` which were being manually set for every `commandDetails` invocation was neat + removing an arg on executeCommands.

## Steps to Reproduce

sample commands:
- run `/ping`
- run `/info`

## Demonstration of Changes
- no edited tags / janky output with embeds

image
image

This PR kinda makes `messageWhileExecuting` redundant, maybe worth removing that?


src/commands/miscellaneous/ping.ts
```diff
@@ -1,42 +1,22 @@
...
-    const apiLatency = getApiLatency(initialMessageFromBot, messageFromUserAsMessage);
-    return new Promise((resolve, _reject) => resolve(content(botLatency, apiLatency)));
-  }
-  const initialPing = await messageFromUserAsMessage.channel.send(initialPingContent);
-  const apiLatency = getApiLatency(initialPing, messageFromUserAsMessage);
-  initialPing.edit(content(botLatency, apiLatency));
-  return Promise.resolve('');
+  const apiLatency = Date.now() - messageFromUser.createdTimestamp;
+  return `${content(botLatency, apiLatency)}`;
 };
```
Picowchew, line 20: The sole exception for an edited message should be the `ping` command (i.e. even the regular command edits the message for `ping`). The previous implementation makes sense since it is "precisely" the time it takes for the bot to respond, which is what we want to measure.

Picowchew (changes requested):
Left a comment with regard to the `ping` command.

src/commands/miscellaneous/ping.ts
hydrobeam: Fair point! I'll rework the impl

hydrobeam: Done in [60c2b16](https://github.com/uwcsc/codeybot/pull/329/commits/60c2b16f86a7de34bd06d70fa202223a0aedf3b5)

src/commands/miscellaneous/ping.ts
```diff
@@ -3,45 +3,44 @@ import { Message } from 'discord.js';
...
+    content: 'Pong?',
+    ephemeral: pingCommandDetails.isCommandResponseEphemeral,
+    fetchReply: true,
+  });
+
+  const apiLatency = initialResponse.createdTimestamp - messageFromUser.createdTimestamp;
+  const stringReply = content(botLatency, apiLatency);
+
+  // have to modify the original interaction when using slash commands
+  // because you cannot edit ephemeral responses.
```
Picowchew, line 31: What does this comment mean exactly? Aren't you currently editing the ephemeral response?

hydrobeam: I am editing it, but it's happening from the interaction itself, not from the response

mcpenguin: I'm confused what you mean by this - what do you mean that the editing is coming from the interaction and not from the response?

hydrobeam: instead of editing the reply that we're sending (`initialRespone`) directly via `initialResponse.edit`, we instead have to use the `messageFromUser` (a slash command `ChatInputInteraction`) and call the`editReply` method on it. We know `messageFromUser` must be a slash command in the second case because only slash commands can proc ephemeral messages
```js
     await messageFromUser.editReply(stringReply);
```

Picowchew: I see. Perhaps do you mind editing the comment to "have to use the original interaction to edit the reply when using slash commands because an ephemeral reply cannot be directly edited" to make it more clear?

hydrobeam: Done in [037362c](https://github.com/uwcsc/codeybot/pull/329/commits/037362c6b847fcb0f9412c923200d7046956bd75)

Picowchew (approved):
LGTM!

Issue #229, opened 2022-07-10
[FEATURE] Set Profile Updates to automatically update discord year roles


smilesthesad:
Relates to https://github.com/uwcsc/codeybot/issues/219
**Describe the problem and the impact.**
Currently adding year roles requires manual work - every year a new role has to manually be created for that year, and the role react message has to be updated to include that year. This ticket attaches years to their profile, so when they set their profile, we can automatically assign roles for them.

**Describe your proposed solution(s).**

In the setProfile command make the following changes:
 

- [x] if they update their year, go through existing validation. If the year is in the currentYears range (curryear -2 - curryear + 7), then check if that role exists. If it doesn't create a role for it. Then, add this year to their profile
- [x] If the year they set is less than the current year, automatically update the alumni role as well 

**Deadlines**
- By next term (Sept 1st)

**Additional Information**
Validation is already done for set profiles, so no malicious user can create an arbitrary year in their profile causing an additional role being created (we have a max of 250 roles)

PR #331, opened 2022-08-18
automate year and alumni roles

smilesthesad:
## Summary of Changes

- When users use .set profile year, year roles will be automatically assigned/removed.
 
## Motivation and Explanation

- Manually creating new roles for each new year is tiring and could be automated. Furthermore, we have an easy way to identify alumni since the profile feature already exists.

## Related Issues

resolves <https://github.com/uwcsc/codeybot/issues/229>
## Steps to Reproduce

- Run .profile set <year> 
- Observe if the year is less than the current year, then you will automatically be assigned the alumni role 
- Observe that if the year is within the currentYears range identified by the original ticket known as the enum "validRoleYears" on the pr, then you will get that year assigned as a role to you.
- Observe that if the year is within the currentYears range and the role does not exist, the bot will automatically create this role.

## Demonstration of Changes

BEFORE: 
image
MESSAGE SENT: 
image
AFTER: 
image

### DIFF
--- a/src/components/profile.ts
+++ b/src/components/profile.ts
@@ -1,6 +1,5 @@
 import { openDB } from './db';
-import { container } from '@sapphire/framework';
-import { ColorResolvable, Message } from "discord.js";
+import { ColorResolvable, Message } from 'discord.js';
 import { getUserFromMessage } from '../codeyCommand';
 
 export interface UserProfile {
@@ -32,14 +31,14 @@ const validTerms: string[] = [
   'Professor',
 ];
 const yearStart = 1900;
-const currentYear = new Date().getFullYear()
+const currentYear = new Date().getFullYear();
 // the last valid year of graduation in the future. while most people graduate in <5 years, +2 years just to be safe
 const yearEnd = currentYear + 7;
 // range of years that can be assigned as a role.
 enum validRoleYears {
   validRoleYearStart = currentYear - 2,
   validRoleYearEnd = yearEnd
-};
+}
 
 export enum configMaps {
   aboutme = 'about_me',
@@ -170,7 +169,7 @@ export const getUserProfileById = async (userId: string): Promise<UserProfile | 
 export const editUserProfileById = async (userId: string, data: UserProfile, message: Message): Promise<void> => {
   // grab oldYear as that might be used later on when updating year roles
   const oldProfileDetails: UserProfile | undefined = await getUserProfileById(userId);
-  let oldYear = oldProfileDetails?.year || null; 
+  const oldYear = oldProfileDetails?.year || null;
 
   const db = await openDB();
   // check if a user exists in the user_profile_table already
@@ -182,7 +181,7 @@ export const editUserProfileById = async (userId: string, data: UserProfile, mes
   let query;
   let onlyCustomization; // declare on outerscope, we need to see if this is the year customization
   let onlyDescription;
-  
+
   if (user.found === 1) {
     for (const [customization, description] of Object.entries(data)) {
       // grab the only customization and its corresponding description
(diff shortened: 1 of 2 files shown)


src/components/profile.ts
```diff
@@ -142,15 +151,20 @@ export const getUserProfileById = async (userId: string): Promise<UserProfile |
 
 // NOTE: data will always be just one customization, but we do not know which one, meaning many of
 // the loops are just to extract an unknown key-val pair
-export const editUserProfileById = async (userId: string, data: UserProfile): Promise<void> => {
+export const editUserProfileById = async (userId: string, data: UserProfile, message: Message): Promise<void> => {
+  // grab oldYear as that might be used later on when updating year roles
+  const oldProfileDetails: UserProfile | undefined = await getUserProfileById(userId);
+  const oldYear = oldProfileDetails?.year || null;
```
Picowchew, line 157: Can this be moved into `updateProfileRoles`, so `oldYear` also doesn't have to be a parameter?

src/components/profile.ts
```diff
@@ -176,4 +190,77 @@ export const editUserProfileById = async (userId: string, data: UserProfile): Pr
...
+
+const addOrRemove = {
+  add: true,
+  remove: false
+};
+
+const updateRoleByRoleName = async (message: Message, roleName: string, add: boolean): Promise<void> => {
+  const member = message.member;
+  const userId = getUserFromMessage(message).id;
+  const role = await message.guild?.roles.cache.find((role) => role?.name === roleName);
```
Picowchew, line 208: I don't think `await` is needed?

Picowchew (changes requested):
I left a few comments. To make the code a bit cleaner, when logging errors, can we do `throw new Error('error message here')` instead, and can we remove the logging of non-errors?

src/components/profile.ts
```diff
@@ -176,4 +190,77 @@ export const editUserProfileById = async (userId: string, data: UserProfile): Pr
...
+  if (gradYear >= validRoleYears.validRoleYearStart && gradYear <= validRoleYears.validRoleYearEnd) {
+    // check if role for that year exists, if not add it
+    const newYearRoleName = gradYear.toString();
+    let findRole = message.guild?.roles.cache.find((role) => role?.name === newYearRoleName);
+
+    if (!findRole) {
+      try {
+        // create role object
+        const newRole = {
+          name: newYearRoleName,
```
marko-polo-cheno, line 246: subtract mod 10 for decade role

marko-polo-cheno (changes requested):

src/components/profile.ts
```diff
@@ -142,15 +151,20 @@ export const getUserProfileById = async (userId: string): Promise<UserProfile |
 
 // NOTE: data will always be just one customization, but we do not know which one, meaning many of
 // the loops are just to extract an unknown key-val pair
-export const editUserProfileById = async (userId: string, data: UserProfile): Promise<void> => {
+export const editUserProfileById = async (userId: string, data: UserProfile, message: Message): Promise<void> => {
```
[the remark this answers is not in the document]
Picowchew: Also, I realized that you can replace the `message` parameter with the `member` obtained from this `message`. It looks like you needed the `message` parameter for the `member`, `userId`, and `guild`, but all of these can also be obtained from `message.member`. This would help with potential re-usability of helper functions.

src/components/profile.ts
```diff
@@ -142,38 +150,142 @@ export const getUserProfileById = async (userId: string): Promise<UserProfile |
...
+  const res = await db.get(`SELECT COUNT(*) as found FROM user_profile_table where user_id = ?`, member.id);
   const user = res;
   let query;
+  let onlyCustomization; // declare on outerscope, we need to see if this is the year customization
+  let onlyDescription;
+  for (const [customization, description] of Object.entries(data)) {
+    // grab the only customization and its corresponding description
+    onlyCustomization = customization;
+    onlyDescription = description;
+  }
```
Picowchew, line 165: Can we do `const [customization, description] = Object.entries(data)[0];` instead? I think it would be more clear without a for loop and without the `only` prefix.

src/components/profile.ts
```diff
@@ -142,38 +150,142 @@ export const getUserProfileById = async (userId: string): Promise<UserProfile |
...
     // escape any instances of ' character by placing another ' in front of it by sqlite specifications
     onlyDescription.replace(/'/g, "''");
     query = `UPDATE user_profile_table SET last_updated=CURRENT_DATE, ${onlyCustomization}=? WHERE user_id=?`;
-    await db.run(query, onlyDescription, userId);
+    await db.run(query, onlyDescription, member.id);
   } else {
+    if (onlyCustomization === 'year') {
+      // onlyDescription here would be the new year role
+      await updateProfileRoles(member, onlyDescription);
+    }
```
Picowchew, line 181: This code logic can be moved before `if (user.found === 1) {` to not duplicate.

Picowchew (changes requested):
I left a few more comments. There are also some merge conflicts to be resolved.

src/components/profile.ts
```diff
@@ -142,38 +150,142 @@ export const getUserProfileById = async (userId: string): Promise<UserProfile |
...
+        await member.guild?.roles.create(newRole);
+      } catch (err) {
+        throw new Error(`Failed to create year role for ${newYearRoleName}: ${err}`);
+      }
+    }
+    // assign that role to the user
+    updateMemberRole(member, newYearRoleName, addOrRemove.add);
+    // final step: remove their existing year role
+    // grab oldYear from database - as this call happens before db update
+  } else {
```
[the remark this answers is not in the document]
smilesthesad: This validation happens in the validUserCustomization function already

Picowchew (approved):
LGTM!

src/components/profile.ts
```diff
@@ -157,41 +165,133 @@ export const getUserProfileById = async (userId: string): Promise<UserProfile |
...
+    member.id,
   );
   const user = res;
   let query;
+
+  // grab the only customization and its corresponding description
+  const [customization, description] = Object.entries(data)[0];
+
+  if (customization === 'year') {
+    // onlyDescription here would be the new year role
```
Picowchew, line 182: Replace `onlyDescription` with `description`?

src/components/profile.ts
```diff
@@ -157,41 +165,133 @@ export const getUserProfileById = async (userId: string): Promise<UserProfile |
...
-    }
+    // if customization is year, change year roles as well
     // escape any instances of ' character by placing another ' in front of it by sqlite specifications
-    onlyDescription.replace(/'/g, "''");
-    query = `UPDATE user_profile_table SET last_updated=CURRENT_DATE, ${onlyCustomization}=? WHERE user_id=?`;
-    await db.run(query, onlyDescription, userId);
+    description.replace(/'/g, "''");
+    query = `UPDATE user_profile_table SET last_updated=CURRENT_DATE, ${customization}=? WHERE user_id=?`;
+    await db.run(query, description, member.id);
   } else {
```
marko-polo-cheno, line 192: this else should be like... === 0 right?

marko-polo-cheno: i guess we never have more than 1 tho

marko-polo-cheno (approved):
is there a way to set roles for everyone so far who has done a user profile?

PR #353, opened 2022-08-25
Added validation for env vars

### LINKED ISSUE #37: Validate environment variables on init
Our bot requires several environment variables as part of its configuration. As we add more, we should start validating on initialization so we aren't debugging config problems during command runtime. For example, make sure the app has all the required env vars, and make sure the channels exist for the given channel IDs.

mcpenguin:
## Summary of Changes
- Moved vars.template.json to root of config from dev
- Added validation for environment variables, from vars.template.json

## Motivation and Explanation
Like described in the issue, it's easy to miss config variables - this helps make sure we don't miss any.

## Related Issues

Resolves #37 

## Steps to Reproduce
If you omit an environment variable from your vars, an error is logged with (one of) the variables you missed, and the bot is not initialized.


src/validateEnvVars.ts
```diff
@@ -0,0 +1,10 @@
+import { vars, varsTemplate } from './config';
+
+// Returns whether environment variables are valid
+export const validateEnvironmentVariables = (): void => {
+  for (const key of Object.keys(varsTemplate)) {
+    if (!vars[key]) {
+      throw new Error(`"${key}" is not defined in vars`);
```
Picowchew, line 7: Could `vars` be replaced with `config/${ENV}/vars.json`?

src/bot.ts
```diff
@@ -53,9 +55,9 @@ export const startBot = async (): Promise<void> => {
       initMessageCreate(client, logger, message);
     });
     client.login();
-  } catch (e) {
-    console.log('Bot failure');
-    console.log(e);
+  } catch (e: unknown) {
+    logger.error('Uh oh, something went wrong when initializing the bot!');
+    logger.error(e);
```
Picowchew, line 60: Could these two lines be combined into one `logger.error`, but separated with `\n`?

Picowchew (changes requested):
Could the instructions in `SETUP.md` also be edited, since `vars.template.json` is now moved out of `config/dev/`?

Picowchew (changes requested):

PR #356, opened 2022-08-25
Added profile grad command

### LINKED ISSUE #230: [FEATURE] Add command for pruning roles and adding alumni roles
**Describe the problem and the impact.**
For every year we need to update the alumni or people don't update their roles so we need a way to quickly do so.
We also need to prune the year roles so we need a command that does that easily and swiftly

**Describe your proposed solution(s).**
Making a new command called `/updateyearroles` only available to execs which will prune the roles which are current year - 3 and set the role `alumni` to people who have their graduation year set as last year. This will need to be run at the start of every year.

Example. If the year is 2022, people who's graduation year is is set to 2021 will have the role alumni added to it and the year role 2019 will be deleted and the channel will be archived

**Deadlines**
By the end of the Spring 2022 Term

Picowchew:
## Summary of Changes

- Added the `.profile grad` command
- Code cleanup

## Motivation and Explanation

The `.profile grad` command assigns decade roles, prunes individual year roles, and assigns Alumni role

## Related Issues

Resolves #230 

## Steps to Reproduce

Run `.profile grad`

## Demonstration of Changes

image

Below are some examples of role changes

Before: image After: image

Before: image After: image

Before: image After: image

Before: image After: image (No change)


src/components/profile.ts
```diff
@@ -118,7 +123,7 @@ export const validUserCustomization = (
 ): userCustomization => {
   let parsedDescription = description;
   if (customization !== validatedFields.term) {
-    // convert first letter to capital in case the user doesnt use proper capitalization
+    // convert to lowercase then first letter to capital, in case the user doesn't use proper capitalization
```
mcpenguin, line 126: lowercase -> `lower case`?

src/components/profile.ts
```diff
@@ -288,10 +232,41 @@ const updateMemberGradRoles = async (member: GuildMember, gradYear: number): Pro
...
+    .forEach((role) => {
+      const roleMembers = role.members;
+      const decadeRoleName = yearToDecade(Number(role.name)) + 's';
+      roleMembers.forEach((member) => updateMemberRole(member, decadeRoleName, addOrRemove.add));
+      guild.roles.delete(role);
+    });
+};
+
+export const assignAlumniRole = async (): Promise<void> => {
+  (await container.client.guilds.fetch(TARGET_GUILD_ID)).roles.cache
```
mcpenguin, line 259: We might just want to query `roles` directly, since the cache might not contain the roles. i.e. `roles.filter. ...` rather than `roles.cache.filter ...`

src/utils/roles.ts
```diff
@@ -0,0 +1,47 @@
...
+    throw new Error(`Failed to create the role ${roleName}: ${err}`);
+  }
+};
+
+export const updateMemberRole = async (
+  member: GuildMember,
+  roleName: string,
+  add: boolean,
+): Promise<void> => {
+  let role = member.guild.roles.cache.find((role) => role.name === roleName);
```
mcpenguin, line 27: Again, use `roles.find. ...` rather than `roles.cache.find ...`

mcpenguin (changes requested):
Small comments

src/components/profile.ts
Picowchew: Alright, I fetched the roles.

src/utils/roles.ts
Picowchew: Yup, I also fetched the roles here.

mcpenguin (approved):
LGTM!

Issue #299, opened 2022-08-15
Fix pluralization of words


Picowchew:
In various commands, Codey does not seem to correctly pluralize words. Below is an example, in which `coins` should be `coin`.
![image](https://user-images.githubusercontent.com/58180935/184563071-26096991-f8be-4d63-b1b1-eeec72fe4277.png)
In addition, it would be nice if Codey correctly determines if a word is supposed to be plural, instead of having `(s)`. Below is an example, in which `coin(s)` would preferably be `coins`.
![image](https://user-images.githubusercontent.com/58180935/184563262-5a6fd0a9-973d-467d-8c21-b8412efc6572.png)

PR #391, opened 2022-09-24
Fix word pluralization

smilesthesad:
## Summary of Changes

Fixed pluralization of codey coin. If the user has more than 1 or 0 codey coins, the pluralization will add an s.
## Motivation and Explanation

Better grammar :)
## Related Issues

Resolves <https://github.com/uwcsc/codeybot/issues/299>
## Steps to Reproduce

Any references to codey coins that rely on an amount (e.g: user updated their balance, user wins in the games that award coins) will dynamically set the pluralization.

## Demonstration of Changes

image
image
image

Pluralize helper only works for codey coins right now, and should work with anything that just adds an s for pluralization. Will not work for something like mouse -> mice. This note is also on the docstring

### DIFF
--- a/src/commandDetails/coin/adjust.ts
+++ b/src/commandDetails/coin/adjust.ts
@@ -12,6 +12,7 @@ import {
   UserCoinEvent,
 } from '../../components/coin';
 import { getCoinEmoji } from '../../components/emojis';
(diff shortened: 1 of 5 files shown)


mcpenguin (approved):
LGTM - thank you for putting comments :))

PR #399, opened 2022-10-04
389 split instances

### LINKED ISSUE #389: Split staging and production bots into 2 VM instances
Currently the two staging and prod bot deployments are on the same VM, which is a potential risk if staging somehow takes down prod.

Should pre-emptively split them into 2 before anything happens

alexz429:
Resolves #389 
Also modifies build.yml to correctly deploy to the now separate instances for staging and production.
The new instance required newer versions of docker compose; hence the switch from `docker-compose` which is v1 to `docker compose` which is current. Modified the old instance to use the new calls as well.

once again, PRODUCTION HAS NOT BEEN TESTED UNTIL THE NEXT RELEASE. Proceed with caution.

Steps to reproduce:
tested to work successfully in the new staging instance 
image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,7 +2,7 @@ name: Build
 
 on:
   push:
-    branches: [main]
+    branches: [main, 389-split-instances]
   pull_request:
     branches: [main]
   release:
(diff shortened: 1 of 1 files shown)


mcpenguin (approved):
LGTM!

Issue #301, opened 2022-08-15
Show ties on coin leaderboard


Picowchew:
Currently, if 2 people have the same coin balance, then they will have different positions on the leaderboard. Instead, they should have the same position. For example, if the first 2 people each have 1000 coins, and the next person has 900 coins, then currently, the leaderboard shows positions 1, 2, and 3. However, the leaderboard should show positions 1, 1, and 3 instead.

marko-polo-cheno:
I believe the best approach is to do this with SQL.
Be mindful of time tho. The more ordering, the slower the query!

https://www.sqlshack.com/overview-of-sql-rank-functions/

probro27:
I have created a fork. Please assign me this issue :)

PR #400, opened 2022-10-09
issue 301 coin leaderboard resolved

probro27:
## Summary of Changes
Changed the model of `UserCoinEntry` to add a rank field and updated the SQL query to get the rank of the user. Updated the `getUserIdCurrentCoinPosition` to get the current rank of the user based on the Rank() function of SQL. 

## Related Issues
To resolve the issue #301 

## Demonstration of Changes
image

### DIFF
--- a/src/commandDetails/coin/leaderboard.ts
+++ b/src/commandDetails/coin/leaderboard.ts
@@ -27,7 +27,7 @@ const getCurrentCoinLeaderboardEmbed = async (
   let currentPosition = 0;
 
   const leaderboardArray: string[] = [];
-  let rank = 1;
+  let rank = 0;
   let previousBalance = -1;
   for (let i = 0; i < leaderboard.length && leaderboardArray.length < limit; i++) {
     const userCoinEntry = leaderboard[i];
--- a/src/commandDetails/coin/leaderboard.ts
+++ b/src/commandDetails/coin/leaderboard.ts
@@ -47,7 +47,7 @@ const getCurrentCoinLeaderboardEmbed = async (
       .join('\\_')
       .split('`')
       .join('\\`');
-    const userCoinEntryText = `${leaderboardArray.length + 1}. ${cleanUserTag} - ${
+    const userCoinEntryText = `${userCoinEntry.rank}. ${cleanUserTag} - ${
       userCoinEntry.balance
     } ${getCoinEmoji()}`;
     leaderboardArray.push(userCoinEntryText);
--- a/src/commandDetails/coin/leaderboard.ts
+++ b/src/commandDetails/coin/leaderboard.ts
@@ -27,7 +27,7 @@ const getCurrentCoinLeaderboardEmbed = async (
   let currentPosition = 0;
 
   const leaderboardArray: string[] = [];
-  let rank = 0;
+  let rank = 1;
   let previousBalance = -1;
   for (let i = 0; i < leaderboard.length && leaderboardArray.length < limit; i++) {
     const userCoinEntry = leaderboard[i];
--- a/src/commandDetails/coin/leaderboard.ts
+++ b/src/commandDetails/coin/leaderboard.ts
@@ -9,7 +9,6 @@ import {
 import {
   getCoinBalanceByUserId,
   getCurrentCoinLeaderboard,
-  getUserIdCurrentCoinPosition,
   UserCoinEntry,
 } from '../../components/coin';
 import { getCoinEmoji } from '../../components/emojis';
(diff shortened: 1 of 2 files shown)


mcpenguin (changes requested):
Hi, thanks for doing this! I'm not too familiar with how OVER works in SQLite, but it seems that there's a bug?
image
During local testing, even though I'm ranked 3rd, the "your position" still says I'm first.

probro27:
Oh I'm sorry. Yeah I only tested with 2 users, one me and one test account. That's why I could have missed this. Let me fix it.

probro27:
Should be fine now. 

image

mcpenguin (changes requested):
Hmm, I still see a bug: if we there is an unknown user in the leaderboard (ie if you do <@[id]> and it says "Unknown User"), it is not displayed in the leaderboard, but the rankings still account for them.
image
For example, there is an "unknown user" in 5th place, so the user in 6th place should really be 5th, instead of 6th.

probro27:
Oh didn't know "Unknown user" was a thing. I got it, will work on it.

src/commandDetails/coin/leaderboard.ts
```diff
@@ -47,9 +48,16 @@ const getCurrentCoinLeaderboardEmbed = async (
...
-    const userCoinEntryText = `${leaderboardArray.length + 1}. ${cleanUserTag} - ${
+    const userCoinEntryText = `${rank}. ${cleanUserTag} - ${
       userCoinEntry.balance
     } ${getCoinEmoji()}`;
+    if (userCoinEntry.user_id === currentUserId) {
+      currentPosition = rank;
+    }
+    if (previousBalance !== userCoinEntry.balance) {
+      previousBalance = userCoinEntry.balance;
+      rank = rank + 1;
```
Picowchew, line 59: By this logic, the leaderboard would look something like the following:
```
1. User1 - 400
2. User2 - 300
3. User3 - 300
3. User4 - 300
3. User5 - 200
4. User6 - 100
```

src/components/coin.ts
```diff
@@ -67,6 +67,7 @@ export const coinBonusMap = new Map<BonusType, Bonus>([
 export interface UserCoinEntry {
   user_id: string;
   balance: number;
+  rank: number;
```
Picowchew, line 70: Since rank is now manually computed, this doesn't seem to be needed?

src/components/coin.ts
```diff
@@ -150,32 +151,17 @@ export const getCurrentCoinLeaderboard = async (limit = 10): Promise<UserCoinEnt
   const db = await openDB();
   const res = await db.all(
     `
-      SELECT user_id, balance
+      SELECT user_id, balance,
+      Rank() OVER (ORDER BY balance DESC) AS rank
       FROM user_coin
-      ORDER BY balance DESC
+      ORDER BY rank ASC
```
Picowchew, line 157: Once again, since rank is now manually computed, the previous simpler logic seems to be fine?

Picowchew (changes requested):

src/components/coin.ts
probro27: Fixed it in this commit

src/components/coin.ts
probro27: Fixed it in this commit.

src/commandDetails/coin/leaderboard.ts
probro27: Updated the logic to display this correctly

probro27:
Hope this works @Picowchew

src/commandDetails/coin/leaderboard.ts
```diff
@@ -47,9 +48,16 @@ const getCurrentCoinLeaderboardEmbed = async (
       .join('\\_')
       .split('`')
       .join('\\`');
-    const userCoinEntryText = `${leaderboardArray.length + 1}. ${cleanUserTag} - ${
+    const userCoinEntryText = `${rank}. ${cleanUserTag} - ${
```
Picowchew, line 51: The leaderboard would currently show the person in 1st place to have a rank of 0?

Picowchew (changes requested):
The implementation still looks buggy. It'd be good to run your bot and test your code?

src/commandDetails/coin/leaderboard.ts
probro27: So sorry. I didn't change much last time so I thought I got it correct. Just updated it and checked it.
image
The ranking should be good now. @Picowchew

probro27:
@Picowchew So sorry. I just seem to forget about it. Let me just fix it and update the PR.

Picowchew (approved):
LGTM!

mcpenguin (approved):
LGTM too!

PR #407, opened 2022-11-14
Introduced pdf preview deletion in the resume channel

saad-a-li:
## Summary of Changes
Introduced resume pdf preview deletion when the original message is deleted.

## Motivation and Explanation
Provides a added level of enhancement and privacy to the resume channel.

## Steps to Reproduce
Send a pdf to your bot's channel and delete the message you sent after the preview shows up.


src/events/messageDelete.ts
```diff
@@ -0,0 +1,27 @@
...
+        return;
+    }
+    const db = await openDB();
+    let res: ResumePreview | undefined;
+    res = await db.get('SELECT * FROM resume_preview_info WHERE initial_pdf_id = ?', message.id);
+    if (res) {
+        let previewMessage = await message.channel.messages.fetch(res.preview_id)
+        await previewMessage.delete();
+    }
+    return;
```
Picowchew, line 26: This `return` statement seems to be redundant and can be removed.

Picowchew (changes requested):
Nice clean code that you wrote so far! :)

Currently, if someone with the permission to delete messages deletes the image preview, and then the PDF file gets deleted, then the bot would crash. To solve this issue, if the image preview gets deleted, then the entry in the database can be removed as well.

In addition, we can remove the entry in the database if the PDF file (and consequently, the image preview as well) gets deleted.

The PR is also not passing build due to linting and formatting. Try `yarn lint` and `yarn format`.

src/components/db.ts
```diff
@@ -138,6 +138,17 @@ const initRpsGameInfo = async (db: Database): Promise<void> => {
...
+const initResumePreview = async (db: Database): Promise<void> => {
+  await db.run(
+    `
+      CREATE TABLE IF NOT EXISTS resume_preview_info (
+        initial_pdf_id VARCHAR(255) PRIMARY KEY NOT NULL,
+        preview_id VARCHAR(255) NOT NULL
+      )
+    `,
+  );
+};
```
Picowchew, line 150: Because we now query by `preview_id`, we can create an index on `preview_id`.

Picowchew (changes requested):

src/components/db.ts
```diff
@@ -147,9 +147,7 @@ const initResumePreview = async (db: Database): Promise<void> => {
       )
     `,
   );
-  await db.run(
-    'CREATE INDEX IF NOT EXISTS ix_preview_id ON resume_preview_info (preview_id)',
-  );
+  await db.run('CREATE INDEX IF NOT EXISTS ix_preview_id ON resume_preview_info (preview_id)');
```
Picowchew, line 150: Following how the other indices are named, it looks like the index name should be `ix_resume_preview_info_preview_id`.

Picowchew (approved):
I left one small comment. Other than that, LGTM! Nice work! :)

mcpenguin (approved):
LGTM!

Issue #380, opened 2022-09-10
Add favourite emoji to user profile


Picowchew:
A field for favourite emoji should be added to user profile.

This field can later be incorporated into multiplayer games. For example:

**Picowchew** :panda_face: vs. **mcpenguin** :penguin: 
\<insert game board here\>

probro27:
I can take this up @Picowchew 
Please assign it to me :)

Picowchew:
Done!

PR #416, opened 2022-12-12
380 add emoji to user profile

probro27:
## Summary of Changes
Added the the `profile_emoji` field in the `user_profile_table`. Added validation to support this field and added some field specific logic. 

## Motivation and Explanation
This allows users to set their favourite emoji as part of their profile. 
It can be further used in games along with name as described as one of the use cases in the related issue. 

## Related Issues
Issue: #380 

## Steps to Reproduce
Command to set profile
`.profile set profile_emoji 🍭 `
Then check:
`.profile about <username>`

## Demonstration of Changes
image
image


src/components/profile.ts
```diff
@@ -156,6 +163,15 @@ export const validUserCustomization = (
         return { reason: 'Invalid year. Must be between: ' + yearStart + ' and ' + yearEnd + '.' };
       }
       break;
+    case validatedFields.profile_emoji:
+      console.log('profile emoji: ' + parsedDescription);
```
mcpenguin, line 167: Remove this?

mcpenguin (changes requested):
It seems that you accidentally run `npm install ...`, which created the `package-lock.json` and changed the `yarn.lock` to link to npm modules (instead of yarn modules). You should revert your module installs and doing `yarn add module` instead of `npm install module` instead.

src/components/db.ts
```diff
@@ -114,7 +114,8 @@ const initUserProfileTable = async (db: Database): Promise<void> => {
             last_updated TIMESTAMP NOT NULL DEFAULT CURRENT_DATE,
             faculty VARCHAR(32),
             program VARCHAR(32),
-            specialization VARCHAR(32)
+            specialization VARCHAR(32),
+            profile_emoji VARCHAR(32)
```
mcpenguin, line 118: At the moment, if `user_profile` already exists (without the new profile_emoji column), this will not work. The table would not be created (since it already exists), which means the new column would not be added either.

This is more so an issue with Codey's underlying design - we should make the addition of columns more dynamic when we create our tables. In the meantime, there seems to be some resources here (https://stackoverflow.com/questions/3604310/alter-table-add-column-if-not-exists-in-sqlite#:~:text=Unfortunately%2C%20SQLite%20doesn't%20support,example%20(but%20in%20C%2B%2B) that you could check out for how to resolve this.

mcpenguin (commented):
One more thing.

src/components/db.ts
probro27: Added the logic to check for existence of column

src/components/profile.ts
probro27: Removed it. Sorry I forgot I was checking this there.

probro27:
@mcpenguin Not sure how the files I have not changed have linter failing. Let me know if it is something I can resolve.

src/components/db.ts
```diff
@@ -114,10 +114,20 @@ const initUserProfileTable = async (db: Database): Promise<void> => {
...
         )
         `,
   );
+  const columns = await db.all(
+    `SELECT count(*) > 0
```
mcpenguin, line 123: This currently doesn't work; columns always has 1 element, namely `count(*) > 0`. I think you could just do `SELECT * FROM ...`?

mcpenguin (changes requested):
For the warnings, it seems that our new build script (a previous PR) now checks for unused variables. In particular, it seems like all of these warnings are Sapphire specific. You can put `// eslint-disable-next-line no-unused-vars - Sapphire specific` above each of the warnings to make the linter happy.

src/components/db.ts
probro27: Okay I'll check this out. Thanks

probro27:
Added `eslint-disable-next-line no-unused-vars` but still the linter is throwing an error in local
image

Adding - Sapphire Specific gives this error 

`error    Definition for rule 'no-unused-vars - Sapphire specific' was not found `

mcpenguin:
Oh okay, I think just add instead
`// Sapphire specific:`
`// eslint-disable-next-line no-unused-vars`

probro27:
@mcpenguin I added the eslint command but still the linter is failing. 
Also I saw that linter also fails for classes if they are the only thing exported from a file. 
So I am not sure what our approach should be for this?

probro27:
Seems to solve the issue: `// eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars`

probro27:
@mcpenguin PR should be good to go :))

mcpenguin (approved):
LGTM! Thank you for dealing with the bug and code fixes :))

PR #420, opened 2022-12-24
Fixed position #0 in coin leaderboard

### LINKED ISSUE #419: Some users get #0 in coin leaderboard

Some users have a rank of #0 when doing the coin leaderboard command, which should not happen.

Examples:
- https://discord.com/channels/667823274201448469/762882537692332064/1055683465238040576
- https://discord.com/channels/667823274201448469/762882537692332064/1055631675620597780

## Expected behavior

The rank of the user should be correctly displayed.

Picowchew:
## Summary of Changes

- Fixed the bug where a user who is not on the coin leaderboard would have position #0
- Stylized `CodeyCoin` as `Codey coin`
- Removed the word "current" from the naming related to the coin leaderboard; if more leaderboards are to be added later, then adjectives can be added just for those

## Motivation and Explanation

This fix would let a user see their correct position

## Related Issues

Resolves #419 

## Steps to Reproduce

Try `.coin lb` when you are not on the coin leaderboard

## Demonstration of Changes

Without ties:
image

With ties:
image

### DIFF
--- a/src/components/coin.ts
+++ b/src/components/coin.ts
@@ -146,16 +146,17 @@ export const changeDbCoinBalanceByUserId = async (
 /*
   Get the leaderboard for the current coin amounts.
 */
-export const getCurrentCoinLeaderboard = async (limit = 10): Promise<UserCoinEntry[]> => {
+export const getCoinLeaderboard = async (limit: number, offset = 0): Promise<UserCoinEntry[]> => {
   const db = await openDB();
   const res = await db.all(
     `
       SELECT user_id, balance
       FROM user_coin
       ORDER BY balance DESC
-      LIMIT ?
+      LIMIT ? OFFSET ?
     `,
     limit,
+    offset,
   );
   return res;
 };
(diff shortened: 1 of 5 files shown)


src/components/coin.ts
```diff
@@ -146,16 +146,17 @@ export const changeDbCoinBalanceByUserId = async (
...
-export const getCurrentCoinLeaderboard = async (limit = 10): Promise<UserCoinEntry[]> => {
+export const getCoinLeaderboard = async (limit: number, offset = 0): Promise<UserCoinEntry[]> => {
   const db = await openDB();
   const res = await db.all(
     `
       SELECT user_id, balance
       FROM user_coin
       ORDER BY balance DESC
-      LIMIT ?
+      LIMIT ? OFFSET ?
```
mcpenguin, line 156: Can we split this into two separate lines?

mcpenguin (changes requested):
Looks good, just had a small change.

mcpenguin (approved):
Upon further research, it seems that LIMIT and OFFSET are usually on the same line. So this PR looks good to me now!

Issue #418, opened 2022-12-13
Make insertion of new SQL table columns more dynamic


mcpenguin:
## Description of bug / unexpected behavior

When reviewing https://github.com/uwcsc/codeybot/pull/416, which adds a new column to the user profile table, I realized that the current design of Codey is inflexible when handling new columns to existing tables. If the table (without the new columns) already exists, then the column would not be added (as the table already exists, and all we do is a `CREATE TABLE IF NOT EXISTS`). Similarly, if a column is deleted/renamed, this would not be reflected in any existing tables (without manual SQL queries).

## Expected behavior

We should implement some abstraction so that new columns can be added into existing tables without any manual SQL queries. This should be able to
- create the table if it does not exist
- not alter any existing columns in the original table that are also in the new table
- add any new columns that were not in the original table but are in the new table
- delete any columns that were in the original table but not in the new table

One method is to implement a helper that takes a list of column data and spits out an SQL query that adds any new columns to a table and deletes any columns that are not in it anymore, whilst keeping any existing ones intact.

probro27:
I have an idea of how to implement this and since I worked on #416 I can integrate that code to work with this helper. 
So @mcpenguin  can you assign this to me :)

probro27:
Hey @mcpenguin, let me know if I can take this up. Seems simple enough to implement :)

PR #421, opened 2022-12-24
Added functionality to dynamically add columns

probro27:
## Summary of Changes
Added a helper function `addSQLColumnIfNotExists` in `db.ts` to add a column if it does not exist in the database table. 

## Motivation and Explanation
This will allow developers to use this function and add columns without having to write the entire logic. 

## Related Issues
#418 

## Demonstration of Changes
The profile emoji functionality seems to be working fine.

### DIFF
--- a/src/components/db.ts
+++ b/src/components/db.ts
@@ -119,15 +119,7 @@ const initUserProfileTable = async (db: Database): Promise<void> => {
         )
         `,
   );
-  const columns = await db.all(
-    `SELECT *
-      FROM pragma_table_info('user_profile_table')
-      WHERE name='profile_emoji'
-    `,
-  );
-  if (columns.length == 0) {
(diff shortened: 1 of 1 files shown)


src/components/db.ts
```diff
@@ -148,6 +140,23 @@ const initRpsGameInfo = async (db: Database): Promise<void> => {
   );
 };
 
+const addSQLColumnIfNotExists = async (
```
mcpenguin, line 143: Can we add some documentation for this function? Explain what its purpose is, as well as what the different parameters are. We should specify that `columnDataType` is a SQL data type, and give an example (say "eg `VARCHAR(32)`").

mcpenguin (changes requested):
Looks good, just had a small change request.

src/components/db.ts
probro27: Got it. I'll add a small documentation comment for this.

mcpenguin (approved):
LGTM!

PR #441, opened 2023-02-04
424 explicit codey user error

### LINKED ISSUE #424: Explicit Codey User Friendly Errors
Previously, we used message.reply early if there were validation errors, e.g: "a user id must be provided". However with the migration to sapphire, we're doing `throw new Error` like so 
image

The issue with this is that we have a default message if we just throw a regular Error, instead of that message actually being sent to the user.
image
We should develop a CodeyUserError, to make it very explicit when we're exiting early due to unintended use, which also makes debugging easier. Whenever we want to return an error message to the user, we can do something like `throw new CodeyUserError("a user id must be provided")` instead of message.reply. In the case we have unexpected errors that we do not want to display to a user, such as errors from API calls, throw new Error can still be used, and have the default message be displayed.

## How can the new feature be used?

Instead of message.reply when something goes wrong, we can throw an error while still giving the user an explanation of what went wrong. This makes it very clear from a regular message.reply to improve our codebase.

probro27:
## Summary of Changes
Edited the current returning of messages and errors to return CodeyUserError which is a new class to represent errors that relate to command validation. 

## Motivation and Explanation
It helps differentiate successful messages to failure messages. 

## Related Issues
resolves #424 

## Steps to Reproduce
Need to run all the commands to see if the error received are correct. 

## Demonstration of Changes
image
image

Some place `undefined` has been passed for message in `CodeyUserError`. This is because message has been set during the catch block (as message was not provided in the function itself). 

PS: Added last `if (game)` condition in blackjack.ts, because game has an optional null type and passing it to functions at the end, it needs to be checked. I was getting some linting errors without it.


src/codeyCommand.ts
```diff
@@ -169,7 +170,7 @@ const setCommandOption = (
         )
       );
     default:
-      throw new Error(`Unknown option type.`);
+      throw new CodeyUserError(undefined, `Unknown option type.`);
```
mcpenguin, line 173: Actually, we should just change this to a regular Error, since this isn't an error propagated to the user.

src/codeyCommand.ts
```diff
@@ -334,7 +335,15 @@ export class CodeyCommand extends SapphireCommand {
       }
     } catch (e) {
       logger.error(e);
-      message.reply(commandDetails.messageIfFailure ?? defaultBackendErrorMessage);
+      console.log(e instanceof CodeyUserError);
```
mcpenguin, line 338: remove this debug message?

mcpenguin (changes requested):
Some comments.

probro27:
@mcpenguin should be good to go

mcpenguin (approved):
LGTM!

PR #444, opened 2023-02-07
Slashify User Profile command

### LINKED ISSUE #248: [FEATURE] Slashify user profile
Create slash commands (whilst preserving the regular commands) for user profile.

probro27:
## Summary of Changes
Build slash command for `profile` by making three separate command details and updating the existing profile.ts

## Related Issues
resolves #248 

## Steps to Reproduce
Check out all slash commands for profile

## Demonstration of Changes
image

Set
image

About
image

Grad
image

### DIFF
--- a/src/commandDetails/profile/about.ts
+++ b/src/commandDetails/profile/about.ts
@@ -47,23 +47,23 @@ const profileAboutExecuteCommand: SapphireMessageExecuteType = async (
         // iterate through each of the configurations, prettyProfileDetails making the configuration more readable
         // as opposed to snake case
         // need to cast val to string since addField does not take in numbers
-        profileDisplay.addField(
-          prettyProfileDetails[key as keyof typeof prettyProfileDetails],
-          val.toString(),
-          key !== 'about_me', // since about_me can be long, we dont want to inline it
-        );
+        profileDisplay.addFields({
+          name: prettyProfileDetails[key as keyof typeof prettyProfileDetails],
+          value: val.toString(),
+          inline: key !== 'about_me', // since about_me can be long, we dont want to inline it
+        });
       }
     }
     // add Codey coins onto the fields as well
     const userCoins = (await getCoinBalanceByUserId(user.id))!.toString();
-    profileDisplay.addField('Codey Coins', userCoins, true);
+    profileDisplay.addFields({ name: 'Codey Coins', value: userCoins, inline: true });
     // display last updated last
     if (profileDetails['last_updated']) {
-      profileDisplay.addField(
-        prettyProfileDetails.last_updated,
-        profileDetails['last_updated'],
-        true,
-      );
+      profileDisplay.addFields({
+        name: prettyProfileDetails.last_updated,
+        value: profileDetails['last_updated'],
+        inline: true,
+      });
     }
     return { embeds: [profileDisplay] };
   }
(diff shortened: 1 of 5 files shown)


src/commandDetails/profile/about.ts
```diff
@@ -0,0 +1,92 @@
...
+        profileDisplay.addField(
+          prettyProfileDetails[key as keyof typeof prettyProfileDetails],
+          val.toString(),
+          key !== 'about_me', // since about_me can be long, we dont want to inline it
+        );
+      }
+    }
+    // add Codey coins onto the fields as well
+    const userCoins = (await getCoinBalanceByUserId(user.id))!.toString();
+    profileDisplay.addField('Codey Coins', userCoins, true);
```
mcpenguin, line 59: When testing it seems that `addField` will be deprecated. I think we should change this to `addFields` instead.

src/commandDetails/profile/set.ts
```diff
@@ -0,0 +1,75 @@
...
+  SapphireMessageResponse,
+} from '../../codeyCommand';
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const profileSetExecuteCommand: SapphireMessageExecuteType = async (
+  client,
+  messageFromUser,
+  args,
+): Promise<SapphireMessageResponse> => {
+  const customization = <keyof typeof configMaps>args['customization'];
```
mcpenguin, line 25: I think "customization" should be a dropdown instead of a string field, since otherwise it's hard to tell the different options available. You can look at this for reference: https://discordjs.guide/interactions/select-menus.html#building-and-sending-select-menus

src/commandDetails/profile/set.ts
```diff
@@ -0,0 +1,75 @@
...
+  const { reason, parsedDescription } = validUserCustomization(customization, description);
+  if (reason === 'valid' && messageFromUser.member) {
+    editUserProfile(<GuildMember>messageFromUser.member, {
+      [configMaps[customization]]: parsedDescription,
+    } as UserProfile);
+    return `${customization} has been set!`;
+  }
+  // if reason is not valid the reason will be returned by the validUserCustomization function
+  const messagePrefix = 'Invalid arguments, please try again. Reason: ';
+  throw new CodeyUserError(messageFromUser, messagePrefix + reason);
```
mcpenguin, line 46: Instead of `+`, I think it's better if you use string formatting instead.

mcpenguin (changes requested):
This looks great! Some comments.

src/commandDetails/profile/set.ts
probro27: @mcpenguin Sounds good. I'll figure this out.

mcpenguin (approved):
LGTM!

src/commandDetails/profile/set.ts
mcpenguin: It doesn't seem like this is implemented yet.

mcpenguin (changes requested):
Looks good, but I still think we should change the slash command to have a dropdown for the field instead.

probro27:
@mcpenguin Choices added:
image

src/commandDetails/profile/set.ts
probro27: Implemented

mcpenguin (approved):
LGTM!

Issue #307, opened 2022-08-17
Make coin transfer subcommand


mcpenguin:
## Description of proposed feature

Make a subcommand to transfer coins from your balance to another person's balance.

## How can the new feature be used?

`.coin transfer <username> <amt>` or `.coin t <username> <amt>`

(with the appropriate slash command versions)

botatooo:
I can give a try at implementing this :D

hydrobeam:
I think it shouldn't be one way and should also prompt the receiving person if they wish to accept the trade, just in case someone else doesn't wanna receive coins for any reason

botatooo:
Ooh I could add a prompting option, opinions on that?

Picowchew:
Yup, a prompt and confirmation from the other person sounds good!

mcpenguin:
I agree with @hydrobeam - we should prompt the other user when transferring coins.

botatooo:
How would this be implemented? Pinging the user in chat with some kind of reaction confirmation or dms?

Picowchew:
I would prefer the former, since I think it would look nicer for the entire transfer to occur publicly.

PR #451, opened 2023-03-01
Issue# 307 (no confirmation)

Fan-Yang-284:
## Summary of Changes
Added transfer subcommand to coin. Includes checking for valid coin values (leq than user balance and geq 1) and the recipient is not the user. Currently does not include receiver confirmation, to be implemented.

## Related Issues
Issue #307

## Demonstration of Changes
image
image

Modified files: 
src/commands/coin.ts: added example coin transfer commands
.eslintrc: prettier formatting

New files:
src/commandDetails: transfer feature functionality

### DIFF
--- a/src/commandDetails/coin/transfer.ts
+++ b/src/commandDetails/coin/transfer.ts
@@ -8,8 +8,8 @@ import {
   SapphireMessageResponseWithMetadata,
 } from '../../codeyCommand';
 import {
+  adjustCoinBalanceByUserId,
   getCoinBalanceByUserId,
-  updateCoinBalanceByUserId,
   UserCoinEvent,
 } from '../../components/coin';
 import { getCoinEmoji } from '../../components/emojis';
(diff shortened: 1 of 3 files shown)


src/commandDetails/coin/transfer.ts
```diff
@@ -0,0 +1,122 @@
...
+
+const coinTransferExecuteCommand: SapphireMessageExecuteType = async (
+  client,
+  messageFromUser,
+  args,
+) => {
+  // First mandatory argument is receiving user
+  const receivingUser = <User>args['user'];
+  if (!receivingUser) {
+    throw new Error('please enter a valid user mention or ID for balance transfer.');
```
mcpenguin, line 26: Capitalize the error messages.

src/commandDetails/coin/transfer.ts
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  // Second mandatory argument is amount
+  const amount = <number>args['amount'];
+  if (typeof amount !== 'number') {
```
mcpenguin, line 31: We should also check that amount is an integer. We don't want someone transferring 69.4204204204 coins and making the leaderboard ugly with the decimals.

mcpenguin (changes requested):
Looks mostly good, left some comments.

src/commandDetails/coin/transfer.ts
Fan-Yang-284: These should also be returning a SapphireMessageResponses and not throwing an error right - to notify the user?

mcpenguin (approved):
LGTM! Feel free to merge.

PR #458, opened 2023-03-05
removed console log and added linter rule

### LINKED ISSUE #457: Remove unnecessary console.log + add rule to linter for console.log

We should add a rule for our linter that warns against console.log messages in our code, so that these do not get pushed to production/staging.

mcpenguin:
## Summary of Changes

Removed console.log messages and added a rule for the linter to warn against console.log/error lines in the code.

## Motivation and Explanation

We should not be pushing debug messages to main.

## Related Issues

Resolves #457


Picowchew (approved):
LGTM!

PR #459, opened 2023-03-05
434 leetcode for codey - change problem-id to problem-ID

mcpenguin:
## Summary of Changes

Slash command name fields must be lower case. So changed field 'problem-ID' to 'problem-id'.

(Just a quick fix for the leetcode command PR).


Picowchew (approved):
LGTM!

PR #461, opened 2023-03-08
Coin Transfer Bug Fixes

Fan-Yang-284:
## Summary of Changes
Added sender balance check at time of transfer handling, added enum value `Invalid` for TransferResult.

Also fixes mistake where newReceiverBalance was passed to the pluralize call for newSenderBalance.

## Motivation and Explanation
Fixes bug where sender can start a transfer, lose coins so that they have fewer than the amount to be transferred, but still have the transfer go through. Previously the command only checked if the sender had enough at the start. 

## Related Issues
Issue was closed, but issue#-307

## Demonstration of Changes
image


mcpenguin (approved):
LGTM! Thanks for getting this resolved so quickly :))

Issue #463, opened 2023-03-13
Reduce number of Discord API calls for updating Top 10 Coin Holder


Picowchew:
Currently, people who previously have the `Top 10 Coin Holder` role and will continue to have this role get this role removed and then added back. Instead, we should be limiting the number of Discord API calls and only make the necessary changes.

PR #465, opened 2023-03-14
Efficient API calls

KevinBacabac:
## Summary of Changes
Uses the difference between the current and expected users with codey coin role to update roles.

## Motivation and Explanation
It reduces API calls.

## Related Issues
Resolves https://github.com/uwcsc/codeybot/issues/463

## Demonstration of Changes
I don't have a guild with enough people :eyes: Please test before merging


mcpenguin (approved):
LGTM!

PR #469, opened 2023-04-04
Logging changed to work with codey user error in case of slash commands

probro27:
## Summary of Changes
Fixing CodeyUserError to work with slash commands. Just needed to change the logging method because logging CodeyUserError with message undefined is failing JSON.stringify so only printing the error message in case of CodeyUserError. 

## Motivation and Explanation
Fixing CodeyUserError to work with slash commands. 

## Steps to Reproduce
Run a slash command with some validation error. 
See that the error is shown.

## Demonstration of Changes
image


mcpenguin (approved):
LGTM!

Issue #484, opened 2023-06-13
Remove discriminators from leaderboard


mcpenguin:
## Description of bug / unexpected behavior

With the new Discord update, discriminators (the #xxxx part) are being phased out. As such, we should remove these from the leaderboard.

PR #485, opened 2023-06-14
Issue #484 - Remove discriminators from leaderboard

jman005:
## Summary of Changes
* Removes the fake #0 discriminator from users that have been assigned a Discord user id. Those that still have discriminators will keep them in the leaderboard.

## Motivation and Explanation
* Discord recently started migrating users with discriminators after their username to unique user IDs.

## Related Issues
* Solves #484.
* 
## Steps to Reproduce
* Run .coin leaderboard with a mix of users that have discriminators and those who don't.

### DIFF
--- a/src/commands/games/blackjack.ts
+++ b/src/commands/games/blackjack.ts
@@ -274,6 +274,7 @@ export class GamesBlackjackCommand extends Command {
           // if player has not acted within time limit, consider it as quitting the game
           game = await performGameAction(author.id, BlackjackAction.QUIT);
           message.reply("you didn't act within the time limit, please start another game!");
+          if (game) game.stage = BlackjackStage.DONE;
         }
       }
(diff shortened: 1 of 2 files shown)


src/commands/games/blackjack.ts
```diff
@@ -274,6 +274,7 @@ export class GamesBlackjackCommand extends Command {
           // if player has not acted within time limit, consider it as quitting the game
           game = await performGameAction(author.id, BlackjackAction.QUIT);
           message.reply("you didn't act within the time limit, please start another game!");
+          if (game) game.stage = BlackjackStage.DONE;
```
mcpenguin, line 277: Shoudl this be here?

mcpenguin (changes requested):
Looks good, just a very small change.

PR #486, opened 2023-06-14
Do not match someone with themself if only one user has coffeechat role

### LINKED ISSUE #417: Remove users who have left the server from coffee chat
Currently, users who have left the server with the coffee chat role are still matched with others.

## Expected behavior
This should not happen - coffee chat matching should ignore users who are not in the server.

Fan-Yang-284:
## Summary of Changes
Unable to reproduce bug as described in #417 - added code that prevents a user from being matched with themself in the edge case of only one user having the role.

## Related Issues
Resolves #417 

## Steps to Reproduce
Run `.coffee match` with only one user holding the coffeechat role.

`.coffeechat match` was not working so I removed the `.coffeechat` examples in the command description.


mcpenguin (approved):
LGTM!

Issue #480, opened 2023-06-03
Issues with transfer if waiting too long


mcpenguin:
## Description of bug / unexpected behavior

There was a coin transfer and a user let it sit for about 20mins before accepting it. The transfer failed to change its state to "accepted" or "rejected" but if you tried to accept it, it would do the actual transfer before failing. Since the accept button doesn't go away, the user was able to hit it multiple times to replay the transaction over and over.

## Expected behavior

The transfer should not happen if the coin transfer times out. One solution would be to disable the buttons when the transfer times out.

PR #487, opened 2023-06-28
add timeout to coin transfer, also ping sender and receiver in embed

Fan-Yang-284:
## Summary of Changes
Added timeout to coin transfer (currently 10 minutes), at which point transfer becomes invalid. Also now instead of displaying username in text it creates a UserMention.

## Related Issues
resolves issue #480 

## Steps to Reproduce
In src/components/coin.ts change the value `TransferTimeout` to some smaller value, run `.coin transfer` and wait.

## Demonstration of Changes
image


mcpenguin (approved):
LGTM!

PR #491, opened 2023-08-09
fix undefined error

siumairice:
- realized that using `&&` doesn't work in this context so reverted back to using a ternary operator.

sorry for the trouble :D

### DIFF
--- a/src/components/admin.ts
+++ b/src/components/admin.ts
@@ -1,6 +1,7 @@
 import { GuildMember } from 'discord.js';
 import { vars } from '../config';
 import { logger } from '../logger/default';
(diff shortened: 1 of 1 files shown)


Picowchew (changes requested):
Nice catch! It looks like there are merge conflicts. I'd recommend rebasing on top of the latest commit in the main branch, and only one new commit would be needed. In addition, extra spaces within the strings shouldn't be added, since they would lead to some indentation issues. As a rule of thumb, before submitting a PR, the code would always be tested.

siumairice:
Sounds good then! Thanks for the feedback and sorry for the later response!!

PR #499, opened 2023-10-07
Modify .member

### LINKED ISSUE #498: Modify .member to explain how to become a member
Currently `.member` expects a UW ID to be passed in - modify it to explain the process of becoming a member when no UW ID is passed in instead.

siumairice:
## Summary of Changes
Modified .member to explain the process of becoming a member instead 

## Related Issues
Closes #498

## Steps to Reproduce
Try `.member`


Fan-Yang-284:
Looks great! Just a couple things:
- could you separate the changes for issues 498 and 477
- for the new message maybe just skip the first line and start with the 'Being a CSC member...'
- since it no longer represents invalid usage, maybe set the embed colour to blue (or another colour you think would be suitable)

mcpenguin:
Seems there are linting errors with this right now.

probro27 (approved):
@siumairice LGTM! Thanks for completing this
Squash and merge it whenever you are ready!

Fan-Yang-284 (changes requested):
Really sorry - I thought the new text changes were only relevant to when the user doesn't pass in an ID. Currently both the cases of 
1) no ID being passed in
2) ID being passed in, not a member 

use the `NOT_MEMBER_DESCRIPTION` message. The first case is fine as is, but in the second case we should still explicitly say that they aren't a CSC member as before, with maybe something like this.

```
return new EmbedBuilder().setColor('Red').setTitle(title).setDescription(`You're not a CSC member! ${getEmojiByName('codey_sad')}
  
${NOT_MEMBER_DESCRIPTION}`);
```

probro27:
Moved to #504 
Closing this PR!

Issue #496, opened 2023-09-30
Threads not created automatically for resume reviews


Fan-Yang-284:
Links to error:
https://discord.com/channels/667823274201448469/1064978680960516237/1156367608275279902
https://discord.com/channels/667823274201448469/1064978680960516237/1156666410202841120
https://discord.com/channels/667823274201448469/1064978680960516237/1156734036241297469
... and the resumes after them.

Note that the threads for those resumes were kindly created by a server member.

mcpenguin:
pdfs don't seem to be created automatically either.

Do we have an update on this?

PR #500, opened 2023-10-21
Fix Issue #496 - Threads not created automatically for resume reviews

Fan-Yang-284:
## Summary of Changes
Fixes threads not being created automatically.

The problem was caused by thread names being longer than 100 characters, since the attachment url includes a bunch of parameters that gets returned by the existing regex.

image

## Motivation and Explanation
Adds an additional `.split('?')` call to get only the pdf title. Also checks the resulting length is less than 100 characters, and if not, calls the thread 'Resume' instead.

## Related Issues
Resolves issue #496 

## Demonstration of Changes
image
image


probro27 (approved):
@Fan-Yang-284 LGTM! Thanks for fixing this issue. 
Maybe add a comment in the code explaining why we are doing the `split('?')` but the rest seems good.

Lucien950:
LMAOO TIME TO REOPEN

PR #501, opened 2023-11-20
Allow bot to ban users outside server

davidgan1218:
## Summary of Changes
Allows bot to ban users outside of server even if they have never joined server, previously bot could only ban users already in server.

Note: Ban Message is only sent to banned user if they are already in the server because on Discord direct messages can only be sent to users who are in mutual servers.

### DIFF
--- a/src/commandDetails/admin/ban.ts
+++ b/src/commandDetails/admin/ban.ts
@@ -1,6 +1,6 @@
 import { CodeyUserError } from './../../codeyUserError';
 import { container } from '@sapphire/framework';
-import { PermissionsBitField, User } from 'discord.js';
+import { PermissionsBitField, User, GuildMemberManager } from 'discord.js';
 import {
   CodeyCommandDetails,
   CodeyCommandOptionType,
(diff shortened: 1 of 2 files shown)


src/commandDetails/admin/ban.ts
```diff
@@ -31,11 +31,9 @@ const banExecuteCommand: SapphireMessageExecuteType = async (client, messageFrom
...
     const days = <number>args['days'];
-    // Get the GuildMember object corresponding to the user in the guild
-    // This is needed because we can only ban GuildMembers, not Users
+    // get Guild object corresponding to server
     const guild = await client.guilds.fetch(vars.TARGET_GUILD_ID);
-    const memberInGuild = await guild.members.fetch({ user });
-    if (await banUser(memberInGuild, reason, days)) {
+    if (await banUser(guild, user, reason, days)) {
       return `Successfully banned user ${user.tag} (id: ${user.id}) ${
         days ? `and deleted their messages in the past ${days} days ` : ``
```
Fan-Yang-284, line 38: Here, just update `days` to `${pluralize('day', days)}` like in `src/components/admin.ts/makeBanMessage`

Fan-Yang-284 (changes requested):
Just before merging can you push another commit for a small tweak that we missed before this but might as well fix now - otherwise LGTM!

src/components/admin.ts
```diff
@@ -9,7 +9,7 @@ const MOD_USER_ID_FOR_BAN_APPEAL: string = vars.MOD_USER_ID_FOR_BAN_APPEAL;
 const makeBanMessage = (reason: string, days?: number): string =>
   `
 Uh oh, you have been banned from the UW Computer Science Club server ${
-    days && `and your messages in the past ${days} ${pluralize('day', days)} have been deleted`
+    days && `and your messages in the past ${days} ${pluralize('day', days)} have been deleted `
```
Fan-Yang-284, line 12: Edit: It looks like this won't work as intended if `days` is undefined, so just revert it back to the ternary operator

probro27 (dismissed):
@davidgan1218 LGTM! Please do make the changes that Fan mentioned and there are some merge conflicts so please resolve them.

Fan-Yang-284:
There still seem to be a few linter errors (pictured below) so just fix those and it'll be good!

image

Fan-Yang-284 (approved):

PR #504, opened 2024-01-18
Explain the membership process when no UW ID is passed in .member

KuroganeToyama:
## Summary of Changes
- Made the uwid field optional instead of required
- An embed with membership info will show up if no uwid is passed in

## Resolved Requests
- Resolved request #498 

## Notes
I'd say my membership explanation isn't exactly the most appealing one, so I'm open to any suggestions to make it more appealing.

### DIFF
--- a/src/commandDetails/miscellaneous/member.ts
+++ b/src/commandDetails/miscellaneous/member.ts
@@ -26,18 +26,22 @@ const getMemberEmbed = async (uwid: UwIdType): Promise<EmbedBuilder> => {
   const title = 'CSC Membership Information';
   const EXPLAIN_MEMBERSHIP = `Oops! You didn't provide a UW ID.
 
-  If you're already a CSC member, please provide a UW ID.
+  Please provide a UW ID to check if you are a CSC member.
 
-  If not, perhaps you're interested in becoming one?
+  If you are not, are you interested in being one?
 
   Being a CSC member comes with gaining access to CSC machines, cloud, email, web hosting, and more! Additional details can be found here! https://csclub.uwaterloo.ca/resources/services/
     
   To sign up, you can follow the instructions here! https://csclub.uwaterloo.ca/get-involved/`;
+  const IS_MEMBER_DESCRIPTION = `You're a CSC member! Hooray! ${getEmojiByName('codey_love')}`;
+  const NOT_MEMBER_DESCRIPTION = `You're not a CSC member! ${getEmojiByName('codey_sad')}
+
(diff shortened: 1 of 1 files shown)


src/commandDetails/miscellaneous/member.ts
```diff
@@ -24,11 +24,20 @@ type UwIdType = string | undefined;
  */
 const getMemberEmbed = async (uwid: UwIdType): Promise<EmbedBuilder> => {
   const title = 'CSC Membership Information';
+  const EXPLAIN_MEMBERSHIP = `Oops! You didn't provide a UW ID.
+
+  If you're already a CSC member, please provide a UW ID.
```
probro27, line 29: Instead of saying that if you are a member provide ID, the message should be something like this - 
`Please provide a UW ID to check if you are a CSC member. If you are not, are you interested in being one ...`

probro27 (changes requested):
@KuroganeToyama The code itself is good. Only change I want is to just edit the message a bit, and take care of the linting issue which can be fixed by putting the entire EmbedBuilder in a single line.

src/commandDetails/miscellaneous/member.ts
```diff
@@ -24,11 +24,20 @@ type UwIdType = string | undefined;
...
+
+  If not, perhaps you're interested in becoming one?
+
+  Being a CSC member comes with gaining access to CSC machines, cloud, email, web hosting, and more! Additional details can be found here! https://csclub.uwaterloo.ca/resources/services/
+    
+  To sign up, you can follow the instructions here! https://csclub.uwaterloo.ca/get-involved/`;
   if (!uwid) {
     return new EmbedBuilder()
       .setColor('Red')
```
probro27, line 38: Also make this embed colour "Blue" because now its not an error when no UW ID is given

src/commandDetails/miscellaneous/member.ts
KuroganeToyama: Hi, I've updated the description of `EXPLAIN_MEMBERSHIP`. I also took care of linting for the `EmbedBuilder`s and reorganized the description constants to make it a bit more organized.

src/commandDetails/miscellaneous/member.ts
KuroganeToyama: I've changed the embed color to blue.

probro27 (approved):
@KuroganeToyama LGTM! Thank you for resolving all of the issues. Whenever you are ready you can squash and merge!

PR #505, opened 2024-01-22
Enabled CodeyBot to nuke message without PDFs in #resume-critique and DM user with explanation

### LINKED ISSUE #432: Nuke messages with no attachments in #resume-critique

#resume-critique should only contain messages with attachments (either pdfs or images). So, any messages without attachments should be removed, and Codey should DM the user with an explanation as to why their message was removed. Something like,

"Hey @username, we've removed your message from #resume-critique since only messages with pdfs or images are allowed to be posted there. If you would like to critique a resume, start a thread on the message containing the resume and write your feedback there!"

KuroganeToyama:
## Summary of Changes
- Imported `channelMention` and `userMention` from `discord.js` for the embed
- Nuke messages without PDFs in #resume-critique 
- DM user with embed explaining why their message got nuked

## Related Issues
- Resolve #432 

## Steps to Reproduce
- Send a message without attachment, with non-PDF attachments, and with PDFs to see the bot in action

## Notes
- I chose to use embeds for the DM because the formatting looks nicer. Still open to any suggestions to make the writing more appealing
- Not sure if we need to use `Logger` for `await message.delete()` and `await client.users.send(user, { embeds: [explainEmbed] })`

### DIFF
--- a/src/events/messageCreate.ts
+++ b/src/events/messageCreate.ts
@@ -17,11 +17,14 @@ import { vars } from '../config';
 import { sendKickEmbed } from '../utils/embeds';
 import { convertPdfToPic } from '../utils/pdfToPic';
 import { openDB } from '../components/db';
+import { spawnSync } from 'child_process';
 
 const ANNOUNCEMENTS_CHANNEL_ID: string = vars.ANNOUNCEMENTS_CHANNEL_ID;
 const RESUME_CHANNEL_ID: string = vars.RESUME_CHANNEL_ID;
 const IRC_USER_ID: string = vars.IRC_USER_ID;
 const PDF_FILE_PATH = 'tmp/resume.pdf';
+const HEIC_FILE_PATH = 'tmp/img.heic';
+const CONVERTED_IMG_PATH = 'tmp/img.jpg';
 
 /*
  * If honeypot is to exist again, then add HONEYPOT_CHANNEL_ID to the config
@@ -92,8 +95,7 @@ const convertResumePdfsIntoImages = async (
   const attachment = message.attachments.first();
   const hasAttachment = attachment;
   const isPDF = attachment && attachment.contentType === 'application/pdf';
-  const isImage =
-    attachment && attachment.contentType && attachment.contentType.startsWith('image');
+  const isImage = attachment && attachment.contentType && attachment.contentType.startsWith('image');
 
   // If no resume pdf is provided, nuke message and DM user about why their message got nuked
   if (!(hasAttachment && (isPDF || isImage))) {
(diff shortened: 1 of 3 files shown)


src/events/messageCreate.ts
```diff
@@ -74,14 +82,39 @@ const punishSpammersAndTrolls = async (
...
 const convertResumePdfsIntoImages = async (
+  client: Client,
   message: Message,
 ): Promise<Message<boolean> | undefined> => {
   const attachment = message.attachments.first();
-  // If no resume pdf is provided, do nothing
-  if (!attachment || attachment.contentType !== 'application/pdf') return;
+
+  // If no resume pdf is provided, nuke message and DM user about why their message got nuked
+  if (!attachment || attachment.contentType !== 'application/pdf') {
```
probro27, line 95: We need to support images as well, since many people upload images directly instead of pdfs. 
Create another condition to handle images and create threads with the same image.

probro27 (changes requested):
@KuroganeToyama Great job with the ticket. Just work on adding the image condition. If there is any issues with supporting certain image formats then we can discuss that.

src/events/messageCreate.ts
KuroganeToyama: I added support for images. It seems to work for most image extension, save for two that I'm not very confident of. If you view my testing channel #di-test-resume, you can see that AVIF images don't get a preview like other extensions. And I couldn't find a sample HEIF image to test for iPhone users.

KuroganeToyama: Actually, I've found sample HEIF images. Same issue with AVIF, no previews.

probro27: Could you try and convert the HEIF and AVIF images into JPG/PNG and then send them into the thread?
Also even if there is an image preview in the message itself, send that in the thread for consistency.

probro27:
@KuroganeToyama Great job! Can you fix the linting errors. Rest looks solid, I'll check it out

src/events/messageCreate.ts
KuroganeToyama: So I managed to convert HEIF/HEIC images to JPG. This was done by using a package that runs shell command to do the conversion. For an HEIF/HEIC image of 2.5 MB, the resulting JPG is 10 MB, which is fine for uploading to Discord. This appears to be common among the converters I tested, and this one is the most efficient one. Then again, I'm willing to bet that resumes won't break this limit.

I tried some other packages that doesn't require CLI but they all don't work with TS, so I got desperate and opted to the CLI ones, which actually worked.

probro27: That's fine. Its going to be rare anyway for people to upload heif/heic.

probro27 (approved):
@KuroganeToyama LGTM! I'll just wait for @Fan-Yang-284 to check this PR out as well and approve it. Once he has approved we can squash and merge it. Thank you so much for figuring out a way to meet our demands.

Fan-Yang-284 (approved):
LGTM!

PR #507, opened 2024-01-26
Send receiver a dm on CodeyCoin transfer to notify them

### LINKED ISSUE #506: Notify receiver upon CodeyCoin transfer
Related to this [PR](https://github.com/uwcsc/codeybot/pull/487): adding user mentions in an embed does not actually ping the user. This was not realised then because I was only testing transfers to my development bot.

To work around this, we send a DM to the receiving user with a link to the CodeyCoin transfer embed.

Fan-Yang-284:
## Summary of Changes
Upon transfers being initiated, send the receiving user a DM with a link to the transfer in order to notify them.

Also edits the transfer embed status message upon transfer completion to display user mentions instead of username

## Motivation and Explanation
We previously added a timeout to transfers, so without the receiving user being aware of the transfer beforehand it is difficult to react and accept it within the timeout.

## Related Issues
resolves #506 

## Steps to Reproduce
Coordinate with a second person to test transfers, or comment out the code that prevents sending transfers to oneself.

## Demonstration of Changes
image
image

Apologies for the delay in getting this fixed

### DIFF
--- a/src/commandDetails/coin/transfer.ts
+++ b/src/commandDetails/coin/transfer.ts
@@ -1,5 +1,5 @@
-import { container } from '@sapphire/framework';
-import { User } from 'discord.js';
+import { SapphireClient, container } from '@sapphire/framework';
+import { CommandInteraction, Embed, EmbedBuilder, Message, User, userMention } from 'discord.js';
 import {
   CodeyCommandOptionType,
   CodeyCommandDetails,
@@ -10,6 +10,7 @@ import {
 } from '../../codeyCommand';
 import { getCoinBalanceByUserId, transferTracker } from '../../components/coin';
 import { getCoinEmoji } from '../../components/emojis';
+import { receiveMessageOnPort } from 'worker_threads';
 
 const coinTransferExecuteCommand: SapphireMessageExecuteType = async (
   client,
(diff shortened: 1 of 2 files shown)


src/commandDetails/coin/transfer.ts
```diff
@@ -63,12 +63,35 @@ const coinTransferExecuteCommand: SapphireMessageExecuteType = async (
...
+  // Send a dm to the transfer receiver
+  const receivingUser = <User>result.metadata['receiver'];
+  const client = <SapphireClient>result.metadata['_client'];
+
+  const mentionUser = userMention(receivingUser.id);
+
+  // do not attempt sending dms to discord bots
+  if (!receivingUser.bot) {
+    const message = <Message>sentMessage;
+    const messageLink = message.url;
```
probro27, line 83: This is showing up as `undefined` in my test. 
image

src/commandDetails/coin/transfer.ts
```diff
@@ -63,12 +63,35 @@ const coinTransferExecuteCommand: SapphireMessageExecuteType = async (
...
-  // Store the message which the game takes place in the game object
+
+  // Send a dm to the transfer receiver
+  const receivingUser = <User>result.metadata['receiver'];
+  const client = <SapphireClient>result.metadata['_client'];
+
+  const mentionUser = userMention(receivingUser.id);
+
+  // do not attempt sending dms to discord bots
+  if (!receivingUser.bot) {
```
probro27, line 81: If the transfer is to a bot, throw a CodeyUserError since it is an invalid operation

probro27 (changes requested):
@Fan-Yang-284 Couple of changes needed but solid work. Not sure why the message link is not showing up, since it does show up in the PR screenshots. I manually printed it out in the logs and it shows up as undefined, however message does exist (since `message.guild?.name` does show up correctly).

probro27 (approved):
@Fan-Yang-284 LGTM! Works perfectly and should be good to go!
Thanks for working on this.

PR #510, opened 2024-02-06
Auto-generated Wiki Page for Commands

### LINKED ISSUE #186: [FEATURE] sapphire .help
while commando had a pregenerated .help command, sapphire does not. Additionally, it seems providing descriptions of subcommands is either impossible or convoluted within code.

Potentially start making a .help README on the github, and setup a .help command that will DM users the link instead.

KuroganeToyama:
## Summary of Changes
- Added script to auto-generate wiki markdown page for commands by grabbing info in implementation files
- Wiki gets updated every time CodeyBot is relaunched
- Updated wiki link of `help` command 

## Related Issues
- Resolve #186 

## Notes
- Open to any suggestions for better wiki design

### DIFF
--- a/docs/COMMAND-WIKI.md
+++ b/docs/COMMAND-WIKI.md
@@ -305,23 +305,23 @@
 - **Subcommands:** None
 
 # SUGGESTION 
-  ## suggestion 
-  - **Aliases:** suggest
-  - **Description:** Handle suggestion functions.
-  - This command will forward a suggestion to the CSC Discord Mods.     Please note that your suggestion is not anonymous, your Discord username and ID will be recorded.     If you don't want to make a suggestion in public, you could use this command via a DM to Codey instead.
-      **Examples:**
-      ``.suggestion I want a new Discord channel named #hobbies.``
-  - **Options:** 
-      - ``details``: Details of your suggestion
-  - **Subcommands:** ``list``, ``update``, ``create``
+## suggestion 
+- **Aliases:** ``suggest``
+- **Description:** Handle suggestion functions.
+- This command will forward a suggestion to the CSC Discord Mods.     Please note that your suggestion is not anonymous, your Discord username and ID will be recorded.     If you don't want to make a suggestion in public, you could use this command via a DM to Codey instead.
+    **Examples:**
+    ``.suggestion I want a new Discord channel named #hobbies.``
+- **Options:** 
+    - ``details``: Details of your suggestion
+- **Subcommands:** ``list``, ``update``, ``create``
 
 # COFFEE CHAT 
-  ## coffee 
-  - **Aliases:** None
-  - **Description:** Handle coffee chat functions.
-  - **Examples:**
-      ``.coffee match``
-      ``.coffee test 10``
-  - **Options:** None
-  - **Subcommands:** ``match``, ``test``
+## coffee 
+- **Aliases:** None
+- **Description:** Handle coffee chat functions.
+- **Examples:**
+    ``.coffee match``
+    ``.coffee test 10``
+- **Options:** None
+- **Subcommands:** ``match``, ``test``
 
--- a/docs/COMMAND-WIKI.md
+++ b/docs/COMMAND-WIKI.md
@@ -1,0 +1,327 @@
+# ADMIN
+## ban
+- **Aliases:** None
+- **Description:** Ban a user.
+- **Examples:**<br>  `.ban @jeff spam`
+- **Options:** 
+    - ``user``: The user to ban.
+    - ``reason``: The reason why we are banning the user.
+- **Subcommands:** None
+
+# COIN
+## coin
+- **Aliases:** None
+- **Description:** Handle coin functions.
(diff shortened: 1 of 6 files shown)


src/utils/updateWiki.ts
```diff
@@ -0,0 +1,253 @@
+import { statSync } from 'fs';
+import { appendFile, writeFile, readdir, readFile } from 'fs/promises';
+
+// File Paths
+const wikiPath = 'docs/COMMAND-WIKI.md';
+const commandsDir = 'src/commands';
+const commandDetailsDir = 'src/commandDetails';
+
+// Commands with subcommands, to be handled separately
+const compoundCommands = ['coin', 'company', 'interviewer', 'leetcode', 'profile'];
```
probro27, line 10: Can you find a way to automate this? I know its too much to ask, but it would really help, since everytime we create a new command we wouldn't need to update here. Thanks

probro27 (changes requested):
@KuroganeToyama Thanks for working on this, really appreciate you clarifying everything with me.
Couple of functionality issues:
- So the script doesn't really work (?) as in when I added a new command (you can reproduce it by copying `info.ts` files from command and commandDetails to a new file and renaming everything) and it didn't update the wiki when I re-started the bot
- Also if you change the description somewhere, and run the bot, it doesn't change the wiki. 
- Also try and support the case where you have deleted the wiki (think it should be supported by writeFile) but not sure why it wasn't working. 

I am running the bot using `yarn image:build && yarn clean && yarn start` as well as just `yarn start`, but didn't see a change. The function is running since I tried adding a log and it worked. I think the issue is with the `wikiPath` since you are using a relative path
My log using `path.join(__dirname, wikiPath)`: `024-02-07 05:36:28 - info - Updating COMMAND-WIKI.md at /usr/app/dist/src/utils/docs/COMMAND-WIKI.md`

This suggests that the wiki being updated is inside the docker container (maybe it worked for you since you were running it without docker). I would suggest you try and find a way to get the repository location (I am not sure myself how and am looking into it).

src/events/ready.ts
```diff
@@ -45,4 +46,11 @@ export const initReady = (client: Client): void => {
   sendReady(client);
   initCrons(client);
   initEmojis(client);
+  logger.info({
```
probro27, line 49: Move the logs inside the `updateWiki` function (keep this function clean)

docs/COMMAND-WIKI.md
```diff
@@ -305,23 +305,23 @@
...
-## suggestion 
-- **Aliases:** suggest
-- **Description:** Handle suggestion functions.
-- This command will forward a suggestion to the CSC Discord Mods.     Please note that your suggestion is not anonymous, your Discord username and ID will be recorded.     If you don't want to make a suggestion in public, you could use this command via a DM to Codey instead.
-    **Examples:**
-    `.suggestion I want a new Discord channel named #hobbies.`
-- **Options:** 
-    - ``details``: Details of your suggestion
-- **Subcommands:** ``list``, ``update``, ``create``
+  ## suggestion
```
probro27, line 308: Make the spacing consistent with the rest of the commands. This one has extra spaces in the subcommands.

probro27 (changes requested):
@KuroganeToyama Great job with the changes. Couple of nitpicks to just wrap this up!

probro27 (approved):
@KuroganeToyama LGTM! I'll just let @Fan-Yang-284 check this as well and then we can merge it in!

src/utils/updateWiki.ts
```diff
@@ -0,0 +1,272 @@
...
+          const options = codeSection.match(optionsPattern)?.[1].split(', ');
+          const subcommandDetails = codeSection.match(subcommandDetailsPattern)?.[1];
+
+          // Format the info
+          await formatCommandName(name);
+          await formatAliases(aliases);
+          await formatDescription(description);
+          await formatDetailedDescription(detailedDescription);
+          await formatOptions(options);
+          await formatSubcommandDetails(subcommandDetails);
```
Fan-Yang-284, line 180: I think we can move this into an `extractAndFormat` function since it gets repeated a few times

src/utils/updateWiki.ts
```diff
@@ -0,0 +1,272 @@
...
+
+const formatSubcommandName = async (name: string | undefined, basecmd: string) => {
+  let formattedName = '';
+  if (name === undefined) {
+    formattedName = 'None';
+  } else {
+    formattedName = name;
+  }
+  await appendFile(wikiPath, `## ${basecmd} ${formattedName}\n`);
+};
```
Fan-Yang-284, line 62: - Since the functionality doesn't seem to differ too much between Command and Subcommand, maybe we can just get rid of the Subcommand version and for subcommands just also pass in the base name with the subcommand name
- It would be odd to provide details/examples/options etc for a command without a name, so if the command name can't be extracted maybe log an error (maybe print `codeSection`) and just return (from the proposed `extractAndFormat` function below) without appending so the entire command won't appear in the wiki

Fan-Yang-284 (commented):
Awesome work on this! This is actually really nice, I just have a few small suggestions to maybe consider.

Fan-Yang-284 (commented):
Thanks - LGTM!

Issue #471, opened 2023-04-04
Rank in coin leaderboard should be by number of people, not coin balance


Picowchew:
## Description of bug / unexpected behavior

Currently, the rank in the coin leaderboard is by the coin balance; this was an oversight.

## Expected behavior

The rank in the coin leaderboard should be by the number of people.

## How to reproduce the issue

This is apparent when there is a tie in the coin leaderboard.

## Additional media files

https://discord.com/channels/667823274201448469/762882537692332064/1092895606999101551

## Additional comments

In the Discord message above, ranks 2-9 should be 3-10.

PR #514, opened 2024-03-02
Issue# 471

andy586586:
## Summary of Changes
Changed CodeyCoin leaderboard to allow for tie handling (i.e. users with same number of CodeyCoins are ranked the same number on the leaderboard).

- A counter called absoluteCounter was introduced to keep track of the number of users iterated through on the leaderboard, to update the rank accordingly.

## Motivation and Explanation
Previously to this commit, ties in CodeyCoin balances would result in misleading output in the leaderboard. For example, multiple users with the same balance would not be ranked the same position:
image

## Demonstration of Changes
image

This PR fixes issue [471](https://github.com/uwcsc/codeybot/issues/471)

### DIFF
--- a/src/commandDetails/coin/leaderboard.ts
+++ b/src/commandDetails/coin/leaderboard.ts
@@ -70,6 +70,7 @@ const getCoinLeaderboardEmbed = async (
         .join('\\_')
         .split('`')
         .join('\\`');
+      // added a "\\" below in ${rank}\\. ${cleanUserTag} so that Markdown does not automatically increment by 1 each time
       const userCoinEntryText = `${rank}\\. ${cleanUserTag} - ${
         userCoinEntry.balance
       } ${getCoinEmoji()}`;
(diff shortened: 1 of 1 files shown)


src/commandDetails/coin/leaderboard.ts
```diff
@@ -64,7 +70,7 @@ const getCoinLeaderboardEmbed = async (
         .join('\\_')
         .split('`')
         .join('\\`');
-      const userCoinEntryText = `${rank}. ${cleanUserTag} - ${
+      const userCoinEntryText = `${rank}\\. ${cleanUserTag} - ${
```
Fan-Yang-284, line 73: Maybe here leave a comment to explain what the double backslash is doing (ie embeds using Markdown) as a reference, in case we run into anything similar in the future

Fan-Yang-284 (commented):
LGTM! Just see the below suggestion. I'll let @probro27 take a look as well but otherwise nice fix!

probro27:
@andy586586 LGTM! We can merge this in :D
Thanks!

probro27 (approved):

PR #516, opened 2024-04-03
Connect4 Bot AI

davidgan1218:
Previously Codeybot would just play in a random column. Implemented minimax algorithm (current depth of 4) to allow bot to calculate best column to play in and make smart moves.


Fan-Yang-284 (approved):
Thanks for working on this - LGTM!

PR #518, opened 2024-05-15
use user mentions instead of tags in coin leaderboard

### LINKED ISSUE #517: Use user mentions in CodeyCoin Leaderboard instead of usernames
Currently, the CodeyCoin leaderboard uses usernames to display as raw strings, but we should use user mentions (`<@id>`) since they look nicer. Note that since the leaderboard is displayed in an embed, the leaders won't be pinged.

See original suggestion and more information here:
https://discord.com/channels/667823274201448469/762882537692332064/1228578319998521375
https://discord.com/channels/667823274201448469/1007014760761540679/1228578890083991592

hyper-neutrino:
## Summary of Changes
- change the coin leaderboard to display user mentions instead of tags

## Motivation and Explanation
- user mentions look nicer
- allow mentioning users on the leaderboard even if the user cannot be fetched for some reason

## Related Issues
- resolves #517

## Steps to Reproduce
- use `/coin leaderboard`

### Before

image

### After

image

### DIFF
--- a/src/commandDetails/coin/leaderboard.ts
+++ b/src/commandDetails/coin/leaderboard.ts
@@ -1,5 +1,5 @@
 import { container, SapphireClient } from '@sapphire/framework';
-import { EmbedBuilder, User } from 'discord.js';
+import { EmbedBuilder } from 'discord.js';
 import {
   CodeyCommandDetails,
   getUserFromMessage,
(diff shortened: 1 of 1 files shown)


Fan-Yang-284 (approved):
LGTM!

PR #519, opened 2024-05-17
add log message for bans in the notif channel

### LINKED ISSUE #436: Log bans as a new message instead of replies

When bans are done, they should be logged as a new message with the following information:
- The user that was banned
- The mod who banned them
- The reason they were banned
- How many days' worth of messages were purged of the banned user.

hyper-neutrino:
## Summary of Changes
- add a log embed (similar to the one for when spammers are kicked) to when the ban command is used
- the old reply will obviously still show up

## Motivation and Explanation
- currently, bans are only indicated as responses to the command invocation
- logs will make this information more readily available and better formatted

## Related Issues
- resolves #436

## Steps to Reproduce
- use `/ban`

## Demonstration of Changes
image

### DIFF
--- a/src/commandDetails/admin/ban.ts
+++ b/src/commandDetails/admin/ban.ts
@@ -1,14 +1,18 @@
-import { CodeyUserError } from './../../codeyUserError';
 import { container } from '@sapphire/framework';
-import { PermissionsBitField, User } from 'discord.js';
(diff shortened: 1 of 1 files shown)


probro27 (approved):
@hyper-neutrino LGTM!

PR #524, opened 2024-05-24
Fix leaderboard error

hyper-neutrino:
The latest leaderboard change accidentally removed suppression of an error that should not be happening. Instead of breaking when the fetched array is empty, we now stop as soon as our iterator goes beyond the end of the array. This will only happen if not enough entries were fetched to fill the fetch limit, meaning we've run out of entries.

This does not affect prod as there are enough users to avoid this, but this error breaks development (and obviously should be fixed anyway).


Fan-Yang-284 (approved):
Thanks for noticing and fixing this - LGTM!

(mb for also missing this last time)

Issue #523, opened 2024-05-22
Fix infinite money exploit (blackjack)


hyper-neutrino:
Currently, users can start a blackjack game, transfer all of their money to another account, and complete the game. Your balance cannot be negative, so this prevents them from losing money, therefore allowing infinite money to be obtained. This is fairly visible but is better to make impossible regardless.

PR #526, opened 2024-06-04
Patch infinite money glitch by mutexing coin transfers and blackjack

hyper-neutrino:
## Summary of Changes
- currently, users can start a blackjack game, transfer their money away, and since balance cannot be negative, they are protected from losses and can get infinite money for free
- this fixes this by preventing transfers and blackjack games from ongoing concurrently

## Related Issues
- #523

## Steps to Reproduce
- attempt to start a blackjack game while a transfer is pending, or attempt to start a transfer while playing blackjack


probro27 (approved):
@hyper-neutrino LTGM! Thanks for catching this bug. This is a good fix. We are good to go :D

PR #527, opened 2024-06-05
401 add pagination helper

### LINKED ISSUE #401: Add Pagination Helper

We previously wanted to paginate suggestions feature <https://github.com/uwcsc/codeybot/issues/32>, but we should create a general helper, which takes in a list of items, and a pageLimit, can output a paginated version of that list. The user should be able to react with emotes to shift left to right in the pages. 

## How can the new feature be used?

Alongside user suggestions, this can potentially be used on the codeycoin leaderboard, and the upcoming company info command <https://github.com/uwcsc/codeybot/issues/122> .

sagarpatel211:
## Summary of Changes
* This ticket addresses the implementation for an efficient pagination helper function.

## Motivation and Explanation
* This feature improve the bot as it expands functionality, such as displaying long paragraphs of text in a paginated way 

## Related Issues
* Resolves #401

## Steps to Reproduce
* Run the discord bot and checkout to branch 401-add-pagination-helper
* Call `.pg`, `.pagination`, or `.pgtest` in your test channel to try out an example pagination I wrote

## Demonstration of Changes
* A sample demo for using pagination is provided through `src/commands/miscellaneous/pagination-test.ts` and `src/commandDetails/miscellaneous/pagination-test.ts`, which can be removed.

* Let me know if there's any changes I can make or improvements in my code style, documentation, etc.!

### DIFF
--- a/docs/COMMAND-WIKI.md
+++ b/docs/COMMAND-WIKI.md
@@ -258,6 +258,13 @@
     - ``uwid``: The Quest ID of the user.
 - **Subcommands:** None
 
+## pg
+- **Aliases:** `pagination`, `paginationtest`, `pgtest`
+- **Description:** Test the pagination feature.
+- **Examples:**<br>  `.pg`<br>  `.pagination`<br>  `.pagination`<br>  `.pgtest`
+- **Options:** None
+- **Subcommands:** None
+
 ## ping
 - **Aliases:** `pong`
 - **Description:** Ping the bot to see if it is alive. :ping_pong:
--- a/src/commandDetails/miscellaneous/pagination-test.ts
+++ b/src/commandDetails/miscellaneous/pagination-test.ts
@@ -1,0 +1,67 @@
+import { container, SapphireClient } from '@sapphire/framework';
+import { CodeyCommandDetails, getUserFromMessage } from '../../codeyCommand';
+import { EmbedBuilder, ButtonInteraction, ComponentType } from 'discord.js';
+import { PaginationBuilder } from '../../utils/pagination';
+import { Command as SapphireCommand } from '@sapphire/framework';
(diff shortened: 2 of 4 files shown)


probro27 (changes requested):
@sagarpatel211 Looks really good. My only issue with this is that the buttons tend to disappear after a while, leaving the embed stuck on a particular page. 
image
Is there a way to fix this?

hyper-neutrino:
@probro27 if we want to avoid memory leaks we cannot keep the buttons around forever (except in cases where the page can be generated from the page number without needing to store the embeds in RAM, but that seems like a future enhancement). maybe to improve the intentionality we can disable the buttons before quitting the listener so it's more obvious that it was paginated and it was intentionally disabled, but that's an aesthetic decision

probro27 (changes requested):
@sagarpatel211 Based on the comment by Alex, the only thing we need to add is a new function that overloads the existing one, except it takes the entire text or leaderboard data and divides it into pages itself. 
We can keep the current function which is useful for specific page data.
Thanks!

sagarpatel211:
New changes include:
* A default timeout of 300 seconds (5 min).
* A default max page of 25.
* A default max characters per page of 2048.
* Navigator buttons disable for 3 seconds before disappearing for aesthetics.
* A new PaginationBuilderFromText function which allows for large multi-line strings as input, separating pages by newlines or reaching the max character count. Decided to have seperate names for clarity since it's an internal helper function.

Let me know if there's more suggestions. If not, after removing the pagination-test command it can be merged.

davidgan1218:
Only the user who invokes the pagination should be able to see the buttons to change pages. Right now, pressing the buttons as an "outside" user results in a "This interaction failed" message.

PR #528, opened 2024-06-23
improve ban command input

hyper-neutrino:
## Summary of Changes
- currently, the ban command only allows inputting a number of days for purging on ban, which is outdated as full second-by-second granularity is now supporter
- additionally, the ban command will currently experience an internal error if the duration provided exceeds the maximum allowance of 7 days instead of offering a descriptive error message
- this change allows inputting any duration and provides a meaningful error message

## Steps to Reproduce
- use the ban command

## Demonstration of Changes

image
image


davidgan1218:
maybe we should add option to specify ban duration as well as message purging duration.

